PR #37, opened 2024-06-13
feat: Include ipc.jsx file in PyPi Wheel

Ahuge:
When installing the deadline cloud for after effects adaptor from PyPi, there was no included `ipc.jsx` file.
This caused renders to not work.

The repository included instructions on building the ipc.jsx file. This PR includes the ipc.jsx in the `hatch build` command and updates the github `release_publish.yml` action to jsxbundle the clientipc before running hatch build.
The PyPi wheel files will now include the ipc.jsx file
Tested locally
N/A
```
```

N/A
N/A
----

### DIFF
--- a/.github/workflows/release_publish.yml
+++ b/.github/workflows/release_publish.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --upgrade hatch
+      - name: Run JSX Bundle for clientipc
+        run: |
+          python jsxbundler.py --source src/aeipc/ipc.jsx --destination src/deadline/ae_adaptor/clientipc/ipc.jsx
       - name: Build
         run: hatch -v build
       # # See https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi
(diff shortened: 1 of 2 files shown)


.github/workflows/release_publish.yml
```diff
@@ -40,6 +40,9 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --upgrade hatch
+      - name: Run JSX Bundle for clientipc
+        run: |
+          python jsxbundler.py --source src/aeipc/ipc.jsx --destination src/deadline/ae_adaptor/clientipc/ipc.jsx
```
moorec-aws, line 45: Running the jsxbundler.py in the release workflow would only result in the package being properly built when running through this specific workflow. For example, building the package locally or in other environments would fail unless the jsxbundler is run first. 

A [build hook ](https://hatch.pypa.io/1.1/plugins/build-hook/) could be added that runs the jsxbundler.py when `hatch build` is initialized. That way the ipc would always be inclued when the package is built. There is an example of one [here](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/hatch_custom_hook.py)

Ahuge: Ah good idea, I will look into that and revise the PR thanks!

PR #44, opened 2024-06-25
feat: Updates to work towards a GA release

Ahuge:
### What was the problem/requirement? (What/Why)
`deadline-cloud-for-after-effects` is an "alpha release" and during my testing trying to get this up and running on a customer managed fleet on AWS I was have had to make several changes to get it working correctly.

### What was the solution? (How)
Each of these commits are aimed to be mostly separate features or fixes however in my testing some of them do build on others.
- [Implement an environment variable to override AfterFx.exe](https://github.com/aws-deadline/deadline-cloud-for-after-effects/commit/15936483bb9a444e8e99363b76d688bb027a8259) This implements optional environment variable overrides to point to a custom "AfterFx.exe" installation.
- [Ensure we include the compiled jsx file in builds](https://github.com/aws-deadline/deadline-cloud-for-after-effects/commit/26dfbaa9f97d92389e8f99b535818e0da0c611e5) This ensures that pypi wheel builds include the client ipc.jsx library. *Note it does not build that jsxbundle. That needs to be included in your CI build system*
- [Launch AfterFx with the -noui flag for headless](https://github.com/aws-deadline/deadline-cloud-for-after-effects/commit/562b484cd2f32f8e5a20c18908fedb4697d3145c) This was required so that AfterEffects is launching headless instead of with a GUI.
- [Change the AfterFx IPC model to a blocking listen](https://github.com/aws-deadline/deadline-cloud-for-after-effects/commit/01393ceccc8fef9adc51600168b243682c6d55d3) In my testing, locally the `listen` IPC call works great...However on AWS and Windows Server the async calls never seem to trigger. This change does have implications that you will see in the next commit. Because the IPC is blocking on the AfterFx.exe side we cannot call `aerender` to hijack AfterFx.exe and do a render. The render command must come through the IPC now.
- [Execute our render over the IPC not aerender](https://github.com/aws-deadline/deadline-cloud-for-after-effects/commit/2a189930c3751d204cd90c09a7dcc410da16eeeb) We're sending the render command over the IPC because `aerender` does not work anymore with the blocking IPC.
- [Implement a socket IPC waiting system for ae_client](https://github.com/aws-deadline/deadline-cloud-for-after-effects/commit/c2ba52227f1de7dd83834076dd7e3e948bae9042) I was having some scenarios where AfterEffects would take longer than the 15 seconds to first boot, this change allows it to take longer on customer systems that may be configured differently (think EBS performance or AfterEffects versions).

### How was this change tested?
[...]

PR #46, opened 2024-06-26
feat: Include custom Adobe Font files in the submitter job

Ahuge:
When a user uses Adobe Fonts, those custom fonts get installed in their appdata directory and are not included in the job.
The resulting render does not have the correct fonts.
We query the scene and collect all of the custom adobe fonts. These fonts are uploaded as job attachments and the adaptor will install the fonts for the currently running user during the launch of After Effects.
When After Effects is closed, the fonts are uninstalled.
After Effects renders that use custom Adobe Fonts render correctly
I have tested it running a modified version of the standard AfterEffects submitter template with a local session-runner.
These files can be found here: https://gist.github.com/Ahuge/ab8da83c16e5ceadb8683bbde4d9155a
The main change in the template was goofing around with my VirtualEnv as a JobEnvironment

I will deploy this to my own Deadline farm and confirm things there as well.

```
```

Not yet

I suppose?
----

### DIFF
--- a/src/deadline/ae_submitter/data/InitData.jsx
+++ b/src/deadline/ae_submitter/data/InitData.jsx
@@ -87,7 +87,79 @@ function __generateInitData()
                 }
             }
         }
+        var fontReferences = generateFontReferences();
+        logger.debug("Get Fonts: " + key, scriptFileInitDataName);
+        for (var i = 0; i < fontReferences.length; i++) {
+            detectedItemsList.push(fontReferences[i]);
+        }
         dcProperties.jobAttachments.autoDetectedInputFiles.set(detectedItemsList);
+    }
+
+    function _generateFontReferences() {
+        var fontLocations = [];
+        var usedList = app.project.usedFonts;
+        for (var i = 0; i < usedList.length; i++) {
+            var font = usedList[i].font;
+            var fontFamilyName = font.familyName;
+            var familyStyle = font.styleName;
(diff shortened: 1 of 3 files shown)


Ahuge:
Will need a rebase once #44 gets merged in.

src/deadline/ae_submitter/data/InitData.jsx
```diff
@@ -87,9 +87,82 @@ function __generateInitData()
...
+        if (parseInt(app.version[0] + app.version[1]) >= 24) {
+            rawFontPaths = _generateFontReferences();
+        } else {
+            rawFontPaths = _oldGenerateFontReferences();
+        }
+        var _tempFilePath = dcUtil.normPath(Folder.temp.fsName + "/" + "tempFonts");
+        var formattedFontPaths = [];
+        var tempFontPath = new Folder(_tempFilePath);
+        if (!tempFontPath.exists) {
+            tempFontPath.create();
```
eherozhao, line 145: https://extendscript.docsforadobe.dev/file-system-access/folder-object.html#folder-object-constructors will new Folder create the folder if it doesn't exist? Could you mind trying it out?

eherozhao:
Thanks for raising the PR. Since we have merged the new submitter as mainline with custom font support, please feel free to try it out and let us know if you have any feedback. I will close this PR since the tracking commit is outdated. We created a branch called submitter_V1 as an archive if you still want to touch the old code and commits.

PR #48, opened 2024-07-04
fix logic around setting render parameters

yuanmich2:
### What was the problem/requirement?
The adaptor was rendering the entire composition every task.

### What was the solution?
The old code was editing the composition, but not the output module of the render queue item already in the render queue. This PR uses timeSpanStart and timeSpanDuration to properly set the render queue item's frame range.

You can actually distribute render across multiple nodes by having different workers work on different frames.

By running render jobs on an EC2 CMF worker in my AWS account.

I don't know where the tests are.

The change brings the code in line with how the adaptor is documented to work.

no

----

### DIFF
--- a/src/aeipc/ipc.jsx
+++ b/src/aeipc/ipc.jsx
@@ -166,47 +166,48 @@ function _ipcModule() {
         if (project) {
             
             // Get the number of compositions in the project
-            var numComps = project.rootFolder.numItems;
+            var numComps = project.numItems;
             var hasComp = false;
             var comp = null;
 
-            // Loop through each item in the project root folder
-            for (var i = 1; i <= numComps; i++) 
+            var hasRQI = false;
+            for (var j = 1; j <= app.project.renderQueue.numItems; j++)
             {
-            
-                // Get the current item
-                var currentItem = project.rootFolder.item(i);
-                
-                // Check if the item is a composition
-                if (currentItem instanceof CompItem) 
+                if(compName == app.project.renderQueue.item(j).comp.name)
                 {
-                    if(currentItem.name.indexOf(compName) !== -1)
+                    hasRQI = true;
+                    conn.writeln("OK");
+                    break;
+                }
+            } 
+            if (!hasRQI)
+            {
+                // Loop through each item in the project to try to find the correct composition
+                for (var i = 1; i <= numComps; i++) 
+                {
+                
+                    // Get the current item
+                    var currentItem = project.item(i);
+                    
+                    // Check if the item is a composition
+                    if (currentItem instanceof CompItem) 
                     {
-                        hasComp = true;
-                        comp = currentItem;
+                        if(currentItem.name.indexOf(compName) !== -1)
+                        {
+                            hasComp = true;
+                            comp = currentItem;
+                            break;
+                        }
                     }
                 }
-            }
-            if (hasComp)
-            {
-                var hasRQI = false;
-                for (var j = 1; j <= app.project.renderQueue.numItems; j++)
-                {
-                    if(compName == app.project.renderQueue.item(j).comp.name)
-                    {
-                        hasRQI = true;
-                        conn.writeln("OK");
-                    }
-                } 
-                if (!hasRQI)
-                {
+                if (!hasComp) {
                     app.project.renderQueue.items.add(comp);
                     conn.writeln("Created RQI");
                 }
-            
-            }
-            else {
-                conn.writeln("Error: Comp doesn't exist in project");
+                else {
+                    doShutdown = true;
+                    conn.writeln("Error: Comp doesn't exist in project");
+                }
             }
         }
     }
@@ -253,33 +254,25 @@ function _ipcModule() {
         var outputPatternData = data["output_pattern"];
         var outputFormatData = data["output_format"];
         var comp = null;
-        var numComps = app.project.rootFolder.numItems;
-        // Loop through each item in the project root folder
-        for (var i = 1; i <= numComps; i++) 
-        {
-            // Get the current item
-            var currentItem = app.project.rootFolder.item(i);
-            
-            // Check if the item is a composition
-            if (currentItem instanceof CompItem) 
-            {
-                if(currentItem.name.indexOf(compName) !== -1)
-                {
-                    comp = currentItem;
-                }
-            }
-        }
         var RQI = null;
+        var compFound = false;
+        //disable all render queue items except for the first item that matches our 
         for (var j = 1; j <= app.project.renderQueue.numItems; j++)
         {
-            if(compName == app.project.renderQueue.item(j).comp.name)
+            if(compName == app.project.renderQueue.item(j).comp.name && !compFound)
             {
+                comp = app.project.renderQueue.item(j).comp;
                 RQI = app.project.renderQueue.item(j);
                 RQI.render = true;
+                compFound = true;
             }
             else {
                 app.project.renderQueue.item(j).render = false;
             }
+        }
+        if(!compFound){
+            doShutdown = true;
+            _log.error("Error: Unable to find composition in render queue");
         }
         // var RQI = RQI_to_copy.duplicate();
         // var RQI = RQI_to_copy;
--- a/src/aeipc/ipc.jsx
+++ b/src/aeipc/ipc.jsx
@@ -274,14 +274,15 @@ function _ipcModule() {
         {
             if(compName == app.project.renderQueue.item(j).comp.name)
             {
-                var RQI_to_copy = app.project.renderQueue.item(j);
-                RQI_to_copy.render = false;
+                RQI = app.project.renderQueue.item(j);
+                RQI.render = true;
             }
             else {
                 app.project.renderQueue.item(j).render = false;
             }
         }
-        var RQI = RQI_to_copy.duplicate();
+        // var RQI = RQI_to_copy.duplicate();
+        // var RQI = RQI_to_copy;
         // var renderQueueItem = app.project.renderQueue.item(1).outputModule(1).getSettings(); // Assuming you want information for the first render queue item
         // alert(renderQueueItem["Format"]);
         output_file_settings = {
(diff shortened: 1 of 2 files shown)


src/aeipc/ipc.jsx
```diff
@@ -166,47 +166,48 @@ function _ipcModule() {
...
+                
+                    // Get the current item
+                    var currentItem = project.item(i);
+                    
+                    // Check if the item is a composition
+                    if (currentItem instanceof CompItem) 
                     {
-                        hasRQI = true;
-                        conn.writeln("OK");
+                        if(currentItem.name.indexOf(compName) !== -1)
```
rmv: What happens if you have two different comps called "Word" and "AnotherWord"? this might match "Word" with "AnotherWord" also.

src/aeipc/ipc.jsx
```diff
@@ -166,47 +166,48 @@ function _ipcModule() {
...
+                            hasComp = true;
+                            comp = currentItem;
+                            break;
+                        }
                     }
-                } 
-                if (!hasRQI)
-                {
+                }
+                if (!hasComp) {
```
rmv: I think this should be `if (hasComp) {`

src/aeipc/ipc.jsx
```diff
@@ -166,47 +166,48 @@ function _ipcModule() {
...
                     app.project.renderQueue.items.add(comp);
                     conn.writeln("Created RQI");
                 }
-            
-            }
-            else {
-                conn.writeln("Error: Comp doesn't exist in project");
+                else {
+                    doShutdown = true;
```
rmv: This would be safer:
```
conn.writeln("Error: Comp doesn't exist in project");
app.project.close(CloseOptions.DO_NOT_SAVE_CHANGES)
shutdown = true;
```

src/aeipc/ipc.jsx
```diff
@@ -253,35 +254,28 @@ function _ipcModule() {
...
+                RQI.render = true;
+                compFound = true;
             }
             else {
                 app.project.renderQueue.item(j).render = false;
             }
         }
-        var RQI = RQI_to_copy.duplicate();
+        if(!compFound){
+            doShutdown = true;
```
rmv: Same as other doShutdown comment above.

src/aeipc/ipc.jsx
```diff
@@ -294,12 +288,17 @@ function _ipcModule() {
...
+        // var frameRate = comp.frameRate;
+        // var start = frame / frameRate;
+        // var end = (frame + 1) / frameRate; 
+        // var duration = end - start
+
+        //Have to set the start and duration according to the last entry in this forum post: https://community.adobe.com/t5/after-effects-discussions/workareastart-and-workareaduration-behaviour-in-extendscript/m-p/13093413
+        var start = currentFormatToTime("0:00:" + frame, comp.frameRate);
+        var duration = currentFormatToTime("0:00:01", comp.frameRate, true);//assumes a duration of one frame 
 
-        setWorkArea(RQI.comp, start, duration, conn);
```
rmv, line 302: if setWorkArea is not used anymore it should be removed.

src/aeipc/ipc.jsx
```diff
@@ -253,35 +254,28 @@ function _ipcModule() {
...
             else {
                 app.project.renderQueue.item(j).render = false;
             }
         }
-        var RQI = RQI_to_copy.duplicate();
+        if(!compFound){
+            doShutdown = true;
+            _log.error("Error: Unable to find composition in render queue");
+        }
+        // var RQI = RQI_to_copy.duplicate();
```
rmv: Better to clean up these comments.

src/aeipc/ipc.jsx
```diff
@@ -294,12 +288,17 @@ function _ipcModule() {
         RQI.outputModule(1).setSettings(output_file_settings);
 
         // Set start frame and end frame
-        var frameRate = comp.frameRate;
-        var start = frame / frameRate;
-        var end = (frame + 1) / frameRate; 
-        var duration = end - start
+        // var frameRate = comp.frameRate;
```
rmv: Would be good to also clean up this commented out old code.

src/aeipc/ipc.jsx
```diff
@@ -294,12 +288,17 @@ function _ipcModule() {
...
+        // var start = frame / frameRate;
+        // var end = (frame + 1) / frameRate; 
+        // var duration = end - start
+
+        //Have to set the start and duration according to the last entry in this forum post: https://community.adobe.com/t5/after-effects-discussions/workareastart-and-workareaduration-behaviour-in-extendscript/m-p/13093413
```
rmv: This comment does not apply anymore right?

src/aeipc/ipc.jsx
```diff
@@ -294,12 +288,17 @@ function _ipcModule() {
...
+        var start = currentFormatToTime("0:00:" + frame, comp.frameRate);
+        var duration = currentFormatToTime("0:00:01", comp.frameRate, true);//assumes a duration of one frame
```
rmv: I think the timecode string format can be changed as part of the project settings, is there a chance this string might not be valid in some cases? or in other words, should we check what is the current format?

Does switching to CurrentFormatToTime let the dcc handle all the rounding complications that were handled before in setWorkArea?

rmv (commented):
This patch fixes the previous code by using RQI specific attributes to set the rendered span instead of modifying the composition. This simplifies the code and avoids a serious bug.

In another PR, it would be a good idea to add some kind of output job bundle test so that regressions can be caught early.

src/aeipc/ipc.jsx
yuanmich2: This if statement was in the original code. I didn't catch that it was doing an indexOf instead of an ==. I am changing it in a new commit and updating this pr

src/aeipc/ipc.jsx
yuanmich2: removed in next commit

src/aeipc/ipc.jsx
yuanmich2: I agree

src/aeipc/ipc.jsx
yuanmich2: implementing this and the previous comment's suggestion in the next commit

src/aeipc/ipc.jsx
yuanmich2: CurrentFormatToTime does handle all rounding errors for us. When we were using setWorkArea, we were sometimes getting errors that we were putting in invalid values.

I'm going to add a check for the formatting

rmv (approved):
LGTM!

src/aeipc/ipc.jsx
```diff
@@ -253,37 +223,27 @@ function _ipcModule() {
...
-            {
-                if(currentItem.name.indexOf(compName) !== -1)
-                {
-                    comp = currentItem;
-                }
-            }
-        }
         var RQI = null;
+        var compFound = false;
+        //disable all render queue items except for the first item that matches our
```
lucaseck, line 228: `That matches our`? Might be missing the end of the comment here

src/aeipc/ipc.jsx
```diff
@@ -166,81 +166,51 @@ function _ipcModule() {
...
-                if (currentItem instanceof CompItem) 
+                if(compName == app.project.renderQueue.item(j).comp.name)
                 {
-                    if(currentItem.name.indexOf(compName) !== -1)
-                    {
-                        hasComp = true;
-                        comp = currentItem;
-                    }
+                    hasRQI = true;
+                    conn.writeln("OK");
```
lucaseck, line 179: Possibly a debug print statement? If not, can it be made more clear what's Ok? Or is something explicitly checking for OK in the logs

src/aeipc/ipc.jsx
```diff
@@ -166,81 +166,51 @@ function _ipcModule() {
...
+                // Loop through each item in the project to try to find the correct composition
+                for (var i = 1; i <= numComps; i++) 
                 {
-                    if(compName == app.project.renderQueue.item(j).comp.name)
+                
+                    // Get the current item
+                    var currentItem = project.item(i);
+                    
+                    // Check if the item is a composition
+                    if (currentItem instanceof CompItem)
```
lucaseck, line 193: nit: These two conditions can likely be combined to reduce nesting

src/aeipc/ipc.jsx
```diff
@@ -253,37 +223,27 @@ function _ipcModule() {
...
-                }
-            }
-        }
         var RQI = null;
+        var compFound = false;
+        //disable all render queue items except for the first item that matches our 
         for (var j = 1; j <= app.project.renderQueue.numItems; j++)
         {
-            if(compName == app.project.renderQueue.item(j).comp.name)
+            if(compName == app.project.renderQueue.item(j).comp.name && !compFound)
```
lucaseck, line 231: Are comp names guaranteed to be unique?

src/aeipc/ipc.jsx
```diff
@@ -253,37 +223,27 @@ function _ipcModule() {
...
             }
             else {
                 app.project.renderQueue.item(j).render = false;
             }
         }
-        var RQI = RQI_to_copy.duplicate();
-        // var renderQueueItem = app.project.renderQueue.item(1).outputModule(1).getSettings(); // Assuming you want information for the first render queue item
-        // alert(renderQueueItem["Format"]);
+        if(!compFound){
+            _log.error("Error: Unable to find composition in render queue");
```
lucaseck, line 243: nit: Can we print the name of the comp that couldn't be found

src/aeipc/ipc.jsx
yuanmich2: They are not. They have unique IDs though and the new submitter/adaptor I'm working on will use that. Rewriting this version of the adaptor to use IDs would be a huge pain though and I think it's outside the scope of this PR

lucaseck (approved):

PR #51, opened 2024-07-09
Changed the adaptor to search the entire project for matching composition names instead of just the root folder

yuanmich2:
Projects that had compositions in folders were not rendering.

The adaptor was only looping through the root folder when searching for compositions. This PR makes it look through the entire project.

You will now be able to render compositions regardless of if they are organized in a folder or not.

I tested with project files that used folders.

No, I don't know what this is.

No, it does not need to be documented because it just addresses a bug.

No

----

### DIFF
--- a/src/aeipc/ipc.jsx
+++ b/src/aeipc/ipc.jsx
@@ -166,47 +166,48 @@ function _ipcModule() {
         if (project) {
             
             // Get the number of compositions in the project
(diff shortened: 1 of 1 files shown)


yuanmich2:
Merging with https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/48

PR #52, opened 2024-07-18
ci: add codeql analysis

moorec-aws:
We need code scanning on PRs

Add CodeQL Code Scanning for PRs, it will also run every monday on mainline

Adds Code Scanning for PRs

CodeQL passes

No

No

----


moorec-aws:
Closing, this is handled in another upcoming change.

PR #55, opened 2024-10-07
chore: fix hatch linting

eherozhao:
The linting failed in the GitHub Action.
Fixed it by changing sys.__stdout__ to sys.stdout.
Fix the Github action for dependency bot.
```
hatch run linting
deadline-cloud-for-after-effects % hatch run lint
cmd [1] | ruff check .
All checks passed!
cmd [2] | black --check --diff .
All done! ✨ 🍰 ✨
12 files would be left unchanged.
cmd [3] | mypy src
Success: no issues found in 9 source files
```

```
```
`hatch run test` but it only has 1 unrelated test
 
n/a
n/a

----


crowecawcaw (approved):
The PR description doesn't quite match the code the code change, but I'd guess they are equivalent.

mwiebe (approved):

eherozhao:
> The PR description doesn't quite match the code the code change, but I'd guess they are equivalent.

Thanks for pointing it out. I thought I had updated it. Will fix it.

PR #63, opened 2024-10-29
feat:Dockable submitter

eherozhao:
We want to have a submitter with only aerender to avoid issues from adaptor.
Pushed the branch with aerender only dockable submitter
unblock more customers
manual test locally by submitting jobs

----


README.md
```diff
@@ -1,117 +1,70 @@
-# AWS Deadline Cloud for After Effects
-
-[![pypi](https://img.shields.io/pypi/v/deadline-cloud-for-after-effects.svg?style=flat)](https://pypi.python.org/pypi/deadline-cloud-for-after-effects)
-[![python](https://img.shields.io/pypi/pyversions/deadline-cloud-for-after-effects.svg?style=flat)](https://pypi.python.org/pypi/deadline-cloud-for-after-effects)
-[![license](https://img.shields.io/pypi/l/deadline-cloud-for-after-effects.svg?style=flat)](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/LICENSE)
+# AWS Deadline CLoud for After Effects: Dockable Submitter
```
crowecawcaw: What's a "dockable submitter"? Also this repo could contain things besides just a submitter in the future right? The old title might have been good enough.

README.md
```diff
@@ -1,117 +1,70 @@
...
-Alternatively, you can specify the output path for the bundler directly with the `--destination` option. 
-See `python jsxbundler.py --help` for details. This may require elevated permissions.
-
-The default scripts location is at `<AE_INSTALL_LOCATION>/Support Files/Scripts`. For example, on Windows this would be: `C:\Program Files\Adobe\Adobe After Effects 2023\Support Files\Scripts`
+## Compatibility
 
-### Submitter Usage
+1. After Effects 24
+1. Python3.9 or higher; and
+1. Windows or a MacOS operating System
```
crowecawcaw: Nit: "... or Mac"

README.md
```diff
@@ -1,117 +1,70 @@
...
 
-To run the script:
+This submitter requires the ability to write files and send communication over the network in order to function properly.
+By default, After Effects scripts are not allowed to perform these actions. To allow scripts to write files or send communication over a network, edit the following settings within After Effects:
 
-1. Open an After Effects project.
-2. Once inside the project, navigate to the `File` Panel at the top.
-3. Under the `Scripts` sub-menu, select the `DeadlineCloudSubmitter.jsx` script to start it.
+    Windows: Select Edit > Preferences > Scripting & Expressions > select Allow Scripts To Write Files And Access Network.
+    macOS: Select After Effects > Settings > Scripting & Expressions > select Allow Scripts To Write Files And Access Network.
```
crowecawcaw: Should this be part of the install process?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
```diff
@@ -0,0 +1,109 @@
...
+    default: 1
+  - name: OutputFiles
+    type: PATH
+    objectType: FILE
+    dataFlow: OUT
+    userInterface:
+      control: HIDDEN
+      label: Output File(s)
+      groupLabel: Frame Range
+    default: ~\Desktop\output
```
crowecawcaw, line 39: Will this have odd effects on Windows?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
```diff
@@ -0,0 +1,109 @@
...
+      label: Starting Frames
+      groupLabel: Frame Range
+    default: 1-10
+  - name: FrameEnds
+    type: STRING
+    userInterface:
+      control: LINE_EDIT
+      label: Ending Frames
+      groupLabel: Frame Range
+    default: 1-10
```
crowecawcaw, line 54: Why are the start and stop defined by frame ranges? I'm used to seeing a single start and stop frame.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
```diff
@@ -0,0 +1,109 @@
...
+  - name: MultiFrameMaxCPU
+    type: INT
+    userInterface:
+      control: SPIN_BOX
+      label: Multi-Frame Max CPU Percentage
+      groupLabel: Source
+    description: The maximum percentage of the CPU Multi-Frame Rendering is allowed to use. Set to 0 to disable Multi-Frame Rendering. Max is 100.
+    default: 90
```
crowecawcaw: From the README, this doesn't work right? Might be better not to include it.

src/OpenAESubmitter.jsx
```diff
@@ -0,0 +1,72 @@
...
+if (isSecurityPrefSet()) {
+    buildUI(this);
+} else {
+    //Print an error message and instructions for changing security preferences
+    var submitterPanel =
+        thisObj instanceof Panel ?
+        thisObj :
+        new Window(
+            "palette",
+            "Submit Queue to AWS Deadline Cloud",
```
crowecawcaw, line 24: Could you share a screenshot of what this looks like? Do users add work to an AE queue and then submit the whole queue to Deadline? The window title seems confusing to me.

src/OpenAESubmitter.jsx
```diff
@@ -0,0 +1,72 @@
...
+            undefined, {
+                resizable: true,
+                closeButton: true,
+            }
```
crowecawcaw, line 28: Super nit: formatting. Would be nice to run this through prettier if it doesn't choke on the `#include` statements.

README.md
eherozhao: That makes sense. Will update.

README.md
eherozhao: Makes sense. Will make the change.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
eherozhao: Yea, sounds reasonable. I can make a change to remove everything related to this feature from this submitter now.

src/OpenAESubmitter.jsx
eherozhao: Yea, will update the wording. I tried to make it both dockable and undockable so customers can choose the way they want. This will only show when it is undockable but the README we have and recommend is the dockable way. 
image

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
eherozhao: should not. We have Artist used the tool in Windows before. Will take an action to improve this.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
eherozhao: This one works for separate frames into tasks for image sequence. I will create an Action Item to validate and improve it.

src/OpenAESubmitter.jsx
eherozhao: yea, I used a formatter but not prettier since this is extendScript. It has a specific plugin for ExtendScript https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/DEVELOPMENT.md so I just format the files with this.

crowecawcaw (commented):
Code looks good to me. I don't have AE set up and haven't tested it manually, it'd be good if the second reviewer had more context on it.

Your strategy of merging a stripped down submitter which works and then adding back features that might be missing makes sense to me. It gives us a working submitter quicker to unblock people, and it lets us develop from a codebase that we understand. It might help to make a list of the things that will be dropped here including user-facing features and ops stuff like `CONTRIBUTING.md` and the GitHub actions so we can make sure to add them back later.

eherozhao:
> Code looks good to me. I don't have AE set up and haven't tested it manually, it'd be good if the second reviewer had more context on it.
> 
> Your strategy of merging a stripped down submitter which works and then adding back features that might be missing makes sense to me. It gives us a working submitter quicker to unblock people, and it lets us develop from a codebase that we understand. It might help to make a list of the things that will be dropped here including user-facing features and ops stuff like `CONTRIBUTING.md` and the GitHub actions so we can make sure to add them back later.

Thanks for the feedback. Yea, I have created internal tickets to track those action items.

src/OpenAESubmitter.jsx
```diff
@@ -0,0 +1,72 @@
+#include "utils/Logger.jsx"
+
+// Setup logger
+var _scriptFileName = "OpenAeSubmitter.jsx";
+var logFileName = "aftereffects" + ".log";
```
eng-josh: Why not all one string? Is there a variable missing here that should have been included in the string concat?

src/OpenAESubmitter.jsx
```diff
@@ -0,0 +1,72 @@
...
+var logDirectoryPath = dcUtil.getUserDirectory() + "/.deadline/logs/submitters/";
+logDirectoryPath = dcUtil.normPath(logDirectoryPath)
```
eng-josh: Nit: Even though `const` technically doesn't exist in Adobe ExtendScript language, we should declare a new variable as opposed to using reassignment.

eng-josh (approved):
Overall, the code looks like it will run fine. Do we need to check in with Doc Writers are on any of the language for messages we may be displaying the customers?

src/OpenAESubmitter.jsx
eherozhao: Makes sense. I will make the update.

src/OpenAESubmitter.jsx
eherozhao: Will make the change.

eherozhao:
> Overall, the code looks like it will run fine. Do we need to check in with Doc Writers are on any of the language for messages we may be displaying the customers?

Yes, before we change the status to Stable from Alpha, I think we will need to have Doc Writers' review on our README.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/aerender.ps1
```diff
@@ -0,0 +1,36 @@
...
+    }
+    if(($output) -match "^aerender ERROR"){
+        exit 2
+    }
+}
+
+$renderarg = @("-project", "`"$project`"", "-rqindex", $rqindex, "-s", $start, "-e", $end, "-v", "ERRORS_AND_PROGRESS", "-close", "DO_NOT_SAVE_CHANGES", "-sound", "OFF")
+
+if (-Not "$outputpath".Contains(",")) {
+    $renderarg += "-output", "`"$outputpath`""
```
npmacl, line 34: I would think a user could append additional arguments by making the value of outputpath something like
`path/to/output --aditional-param1 additional_value1 --additional-param2 additional_value2`. Is there anything a user could do with this we should care about?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
```diff
@@ -0,0 +1,100 @@
...
+    type: INT
+    userInterface:
+      control: SPIN_BOX
+      label: Render Queue Index
+      groupLabel: Source
+    description: The index of the item in the render queue to render.
+    default: 1
+  - name: OutputFiles
+    type: PATH
+    objectType: FILE
```
npmacl, line 33: The name here is plural, but the objectType of this path is `FILE` which I would think means this variable represents a single output file that would be generated. If this is just a single file, could we update the name to be singular?

src/OpenAESubmitter.jsx
```diff
@@ -0,0 +1,72 @@
...
+    root.alignChildren = ["fill", "top"];
+    var errorText = root.add("statictext", undefined, "", {
+        multiline: true,
+    });
+    errorText.graphics.foregroundColor = errorText.graphics.newPen(
+        errorText.graphics.PenType.SOLID_COLOR,
+        [1.0, 0.2, 0.2],
+        1
+    );
+    errorText.text = "⚠ ERROR: Insufficient Script Permissions ⚠";
```
npmacl: does ⚠ render correctly in worker agent logs?

eherozhao: This will not be logged in worker agent side. This will just print out in AE window, like this.
image

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/aerender.ps1
eherozhao: Not really, from customer's perspective, they should not touch this PS file. The passed in outputpath is auto detected from the submitter so the proposed case should not appear.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.yaml
eherozhao: Updated to singler. We can add plural back when supporting multiple output modules.

npmacl (approved):

eng-josh (approved):

Ahuge:
Are we able to modify the job template so that it calls the Step the name of the Comp that it is rendering?
Similar to the adaptor workflow?

New submitter
image


Old submitter
image

eherozhao:
> Are we able to modify the job template so that it calls the Step the name of the Comp that it is rendering? Similar to the adaptor workflow?
> 
> New submitter image
> 
> Old submitter image

Thanks for the comment. Yes, I can look into it.

src/OpenAESubmitter.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+function isSecurityPrefSet() {
+    var securitySetting = app.preferences.getPrefAsLong(
+        "Main Pref Section",
+        "Pref_SCRIPTING_FILE_NETWORK_SECURITY"
+    );
+    return securitySetting == 1;
+}
+
+#include "UI/SubmitterUI.jsx"
```
viknith: Any reason why we should stagger the imports like this where there's logger first, then a batch of imports, then the submitter UI at the end?

src/submission/JobTemplateHelper.jsx
```diff
@@ -0,0 +1,154 @@
+/**
+ * Generates the basic paramterValue file for the job template
```
viknith: Nit: parameter

src/ui/SubmitterUI.jsx
```diff
@@ -0,0 +1,163 @@
...
+    framesPerTaskValue.enabled = separateFramesCheckbox.value;
+
+    separateFramesCheckbox.onClick = function() {
+        app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_SEPARATEFRAMESINTOTASKS, separateFramesCheckbox.value.toString())
+        framesPerTaskValue.enabled = separateFramesCheckbox.value;
+    }
+
+    var submitButton = controlsGroup.add("button", undefined, "Submit");
+    submitButton.onClick = function() {
+        SubmitSelection(list.selection, separateFramesCheckbox.value ? parseInt(framesPerTaskValue.text) : 1);
```
viknith, line 86: Nit: I know list gets set because of the updateList() call further below, but should we consider a null check on list here if it's possible for list.selection to result in an error?

src/submission/SubmitBundle.jsx
```diff
@@ -0,0 +1,174 @@
+/**
+ * Submit the selected render queue item
+ **/
+function SubmitSelection(selection, framesPerTask) {
+    // first we must verify that our selection is valid
+    if (selection == null) {
+        adcAlert("Error: No selection");
+        return false;
```
viknith: I noticed there are empty `return` statements below but here it's `return false`. Does a return type matter? It doesn't look like it, so as a nit, we'd want to it to be consistent to minimize confusion.

viknith (approved):

src/ui/SubmitterUI.jsx
eherozhao: Yes, we have a null check in the `SubmitSelection` method.

src/OpenAESubmitter.jsx
eherozhao: Yes, the reason is we are having a python script as a build system to combine all jsx files to a giant single file as the final submitter jsx script. In this case, this include will be replaced by the content of the file and we want the logger to be initialized before so that any logger in SubmitterUI can work. I can change the order a bit in next revision.

eng-josh (approved):

PR #72, opened 2024-12-02
feat!: Update AfterEffects environment variable naming

Ahuge:
This now matches the other new DCC naming scheme

The assumption that the executable afterfx.exe exists on the PATH doesn't always make sense, often it is easier to specify the version of AfterEffects to use by explicit path and using an environment variable is an easy way to do this for customers.

We have defined a new environment variable `AFTERFX_EXECUTABLE` which follows the naming from 
[`deadline-cloud-for-cinema-4d`](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py#L298) of `<EXE_NAME>_EXECUTABLE`

Breaking change as we changed the name of the existing environment variable.

```
```

Yes, FAQ.md and DEVELOPMENT.md
Yes
----

### DIFF
--- a/src/deadline/ae_adaptor/AEClient/ae_client.py
+++ b/src/deadline/ae_adaptor/AEClient/ae_client.py
@@ -44,7 +44,7 @@ class AEClient(ClientInterface):
             client_path_abs = client_path_abs.replace("\\", "\\\\")
         startup_script_inline = f"var x = new File('{client_path_abs}') ; x.open(); eval(x.read()); app.exitAfterLaunchAndEval = false;"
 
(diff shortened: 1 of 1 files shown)


eherozhao (approved):

eherozhao:
LGTM. @Ahuge do you mind updating the commit message to be semantic to pass the check?

eherozhao:
Thanks for raising the PR. Since we have merged the new submitter as mainline with environment variable naming, please feel free to try it out and let us know if you have any feedback. I will close this PR since the tracking commit is outdated. We created a branch called submitter_V1 as an archive if you still want to touch the old code.

PR #77, opened 2024-12-11
feat!: Support Frame Chunking for tasks

Ahuge:
### What was the problem/requirement? (What/Why)
There can be a high overhead for launching AfterEffects (aerender) for each frame and for short renders, this overhead can be more than half of the entire render time for the job.

### What was the solution? (How)
This solution uses a clever taskParameterDefinitions range function to implement frame chunking from the `OpenJD-Specifications` wiki: [Search for FrameEndMinusOne](https://github.com/OpenJobDescription/openjd-specifications/wiki/Introduction-to-Creating-a-Job#22421-with-a-combination-expression)

```
"taskParameterDefinitions": [
                {
                    "name": "FrameChunkStart",
                    "type": "INT",
                    "range": "{{Param.FrameStart}}-{{Param.FrameEnd}}:{{Param.ChunkSize}}"
                },
                {
                    "name": "FrameChunkEnd",
                    "type": "INT",
                    "range": "{{Param.FrameStartPlusChunkSizeMinusOne}}-{{Param.FrameEndMinusOne}}:{{Param.ChunkSize}},{{Param.FrameEnd}}"
                }
            ],
            "combination": "(FrameChunkStart, FrameChunkEnd)"
```
### What is the impact of this change?
A new UI field *Frames Per Task* in *Job-Specific Settings*
image

That field now controls the chunk size for each task.
This modifies the JobTemplate parameters with two new "calculated" parameters `FrameStartPlusChunkSizeMinusOne` and `FrameEndMinusOne` which are calculated in the submitter and passed as parameters to the template. 

### How was this change tested?
We ran jobs through the three Comp Submission workflows: "Submit One Comp", "Submit Entire Queue As One Job", "Submit Entire Queue As Separate Jobs". 

### Is this a breaking change?
Yes
----
## Runtime Data

I have some runtime metrics from a naive aerender (1 frame per task) workflow and a chunked aerender workflow:

Naive Submission:
```
 ==== SUMMARY ====

Session Count: 38
Session Total Duration: 0:16:51.511000
Session Action Count: 203
Session Action Total Duration: 0:15:12.578000
Task Run Count: 51
Task Run Total Duration: 0:14:54.141000 (88.4%)
Non-Task Run Count: 152
Non-Task Run Total Duration: 0:00:18.437000 (1.8%)
Sync Job Attachments Count: 0
Sync Job Attachments Total Duration: 0:00:00 (0.0%)
Env Action Count: 152
[...]

Issue #73, opened 2024-12-03
Feature request: Movie File Output


Ahuge:
### Use Case

It is a common use case for After Effects to render out video files (.mov, .avi, etc).

Currently the submitter only supports frame based workflows.

### Proposed Solution

We could look at the `deadline-cloud-for-nuke` submitter workflows and how they handle the mov outputs.
Essentially it will be a modification of the Step so that only a single task is created to render the entire MOV.

dzavada:
Definitely you should always have the ability to select chunk sizes it's way more efficient and faster.

eherozhao:
Thanks for your feedback. Our team is having a RFC about chunking here https://github.com/OpenJobDescription/openjd-specifications/pull/54 feel free to leave your comments.

eherozhao:
We merged the PR and now you can visit `feature_dockable_submitter` to use the submitter and we are working on making it as mainline shortly. Please let us know if you have any feedback on the new submitter.

PR #78, opened 2024-12-11
feat!: Quicktime Movie Submission Workflows

Ahuge:
Current submissions were only supported for Frame based workflows and if you were rendering out a movie file, it would continuously overwrite it. See #73 
Implement an automatic movie workflow if we detect the output of a comp is a movie extension
This will automatically submit a single task job for the movie render if it detects it.
Ran the workflow through all three submission queue types and tested both with a scene containing a frame comp and and a movie comp as well as just two movie comps.
No
Yes, functionality for movie workflows has been changed
----

** This is built on top of #77 and should be on hold until that gets in


eherozhao:
Thanks for the PR! We are having a new dockable submitter with aerender and I think it should automatically solve this issue that old submitter had. Do you want to try it out and see if the new submitter works for this .mov output out of box?

PR #83, opened 2024-12-17
fix: use shell executable to call deadline cli for macOS.

eherozhao:
I noticed that the path of deadline might cause the AE application fail to open the deadline gui submitter.

Borrowed the idea from Keyshot submitter https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/11286d7b9b187a546607daec42ff2d5257b4036f/src/deadline/keyshot_submitter/Submit%20to%20AWS%20Deadline%20Cloud.py#L324 that using shell executable to call deadline which will resolve the problem.

Customers should be able to get deadline cli called correctly from the submitter.

I manually test it with my macOS env. Before this change, the submitter failed to open the gui submitter with deadline not found. After this change, the gui submitter was able to open with no issue.

N/A
N/A
----

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -2094,8 +2094,6 @@ function buildUI(thisObj) {
 }
 
 
-logger.log("Running driver file", _scriptFileName, LOG_LEVEL.INFO);
-
 function isSecurityPrefSet() {
     var securitySetting = app.preferences.getPrefAsLong(
         "Main Pref Section",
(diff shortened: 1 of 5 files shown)


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1396,22 +1396,21 @@ function SubmitSelection(selection, framesPerTask) {
 
     // Runs a bat script that requires extra permissions but will not block the After Effects UI while submitting.
     // The following commented-out line will block the UI until the submission window is closed, but it doesn't require extra permissions
```
viknith: Incorrect comment

src/submission/SubmitBundle.jsx
```diff
@@ -156,21 +156,20 @@ function SubmitSelection(selection, framesPerTask) {
 
     // Runs a bat script that requires extra permissions but will not block the After Effects UI while submitting.
     // The following commented-out line will block the UI until the submission window is closed, but it doesn't require extra permissions
```
viknith: Same comment as above

viknith (approved):

npmacl (approved):

PR #86, opened 2024-12-24
fix: Remove CondaPackages parameter override to restore default param...

viknith:
Whenever I submitted a job from After Effects, the CondaPackages parameter field in the Deadline GUI submitter that pops up would be empty even though I have the CondaPackages parameter defined in my production queue's queue environment. The reason for this is because we are overriding the value for that in the submitter code.

Remove the override.

The rest of the changes are formatting adjustments that occured on my local machine.

Easier job submission experience for After Effects artists that have their queue environment set up so that they don't have to keep typing in "aftereffects" and any other plugins they use.

Ran `python3 jsxbundler.py --source src/OpenAESubmitter.jsx --destination dist/DeadlineCloudSubmitter.jsx` and then loaded up the files into AE and tried submitting a job. This time the conda package field had "aftereffects" in it. Double confirmed this with @eherozhao .

Screenshot 2024-12-24 at 10 49 01 AM

No need

Nope, because the default override was an empty string, so now it's just whatever the customer may have preconfigured or blank if they haven't configured anything.

----

### DIFF
--- a/src/submission/JobTemplateHelper.jsx
+++ b/src/submission/JobTemplateHelper.jsx
@@ -35,10 +35,7 @@ function parameterValues(
     }
 
     return JSON.stringify({
-        parameterValues: [{
-                name: "CondaPackages",
-                value: "",
-            },
+        parameterValues: [
             {
                 name: "deadline:targetTaskRunStatus",
                 value: "READY",
--- a/src/submission/JobTemplateHelper.jsx
+++ b/src/submission/JobTemplateHelper.jsx
@@ -37,6 +37,10 @@ function parameterValues(
     return JSON.stringify({
         parameterValues: [
             {
+                name: "CondaPackages",
+                value: "aftereffects",
+            },
+            {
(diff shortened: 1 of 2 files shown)


src/submission/JobTemplateHelper.jsx
```diff
@@ -35,10 +35,7 @@ function parameterValues(
     }
 
     return JSON.stringify({
-        parameterValues: [{
-                name: "CondaPackages",
-                value: "",
-            },
+        parameterValues: [
```
viknith: This is the important change that has the fix. The rest is formatting.

eherozhao (approved):

eng-josh (approved):

mwiebe (approved):

eherozhao:
One problem of this PR is that CMF customers without using Conda env will have the default Conda package when they submit the job and fail the deploy conda task. I don't really think this is always correct. We should make sure we are not breaking some customers' workflow when trying to support other customers' workflows.

PR #87, opened 2024-12-26
fix: replace hard-coded template.yaml with code generated json file

eherozhao:
We are having a hard-coded template.yaml in the job template folder. It is not as easy as using a json file in submitter development to make the template update very flexible.

Changed to use code generated json job template.

Will be very flexible for future job template operations.

Manual submitted job with mov output and image sequence output.

N/A
N/A
----


src/submission/JobTemplateHelper.jsx
```diff
@@ -152,4 +151,160 @@ function findJobAttachments(rootComp) {
...
+            }
+        }
+    }],
+    "steps": [{
+        "name": "{{COMPNAME}}",
+        "hostRequirements": {
+            "attributes": [{
+                "name": "attr.worker.os.family",
+                "anyOf": [
+                    "windows"
```
viknith: Nit: We could also include macOS here right?

eherozhao: I can add it in the next PR. Pretty low risk, we didn't see any customers using Mac worker yet.

viknith (approved):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1100,6 +1100,162 @@ function findJobAttachments(rootComp) {
...
+            "name": "JobScriptDir",
+            "description": "Directory containing embedded scripts.",
+            "userInterface": {
+                "control": "HIDDEN"
+            },
+            "type": "PATH",
+            "objectType": "DIRECTORY",
+            "dataFlow": "IN",
+            "default": "scripts"
+        }
```
mwiebe: I think it would be nice to add the `CondaPackages` parameter definition here too, to be able to support queues that use it and queues that don't at the same time. The definition could look like this:

```json
        {
            "name": "CondaPackages",
            "type": "STRING",
            "userInterface": {
                "control": "HIDDEN",
            },
            "default": "aftereffects=25",
            "description": "If a queue accepts this parameter, it will create a conda virtual environment from it."
        },
```

and then the dynamic submission code can swap in the appropriate default value (including the correct After Effects version number, and any additional plugins that are needed.

eherozhao: Thanks for the comment! The new commit includes this fix.

yuanmich2:
Can we keep the template in a separate file? The main reason why I made it a separate file in the first place was to make it easier to find and edit. The LSP I was using didn't recognize the .jsx file format, but it was able to help with formatting and highlighting errors in .yaml files. If a user ever wants to add their own pre or post steps, I think it's a lot easier to figure out what to do if there's a separate file that they can edit if they know OpenJD.

I do like the idea of switching from yaml to json to make it easier to make changes through code though. I just don't want to give up the ease of making changes by hand.

eherozhao:
> Can we keep the template in a separate file? The main reason why I made it a separate file in the first place was to make it easier to find and edit. The LSP I was using didn't recognize the .jsx file format, but it was able to help with formatting and highlighting errors in .yaml files. If a user ever wants to add their own pre or post steps, I think it's a lot easier to figure out what to do if there's a separate file that they can edit if they know OpenJD.
> 
> I do like the idea of switching from yaml to json to make it easier to make changes through code though. I just don't want to give up the ease of making changes by hand.

Thanks for your reply! The reason was other dcc integrations are using the same in-code json file. I can definitely move it out as a separate json file. We can also reevaluate which way is better for customers then.

mwiebe (approved):

PR #88, opened 2024-12-28
fix: Use JSON data structure to change job template parameters

eherozhao:
1. The replace method to change job template is not good and flexible.
2. We would like to add a RezPackages support for customers who would build ae as a Rez Package.
1. Used JSON data structure to make job template change.
2. Added a hidden job parameter in the template.json
1. Using JSON data structure can avoid making unexpected string replacement.
2. Customers with Rez AE pacakge and use the template.
Changed the queue environment with a RezPackages parameter and the aftereffects-24 was overridden then.
image

N/A
N/A
----


dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.json
```diff
@@ -91,6 +91,15 @@
...
             "description": "If a queue accepts this parameter, it will create a conda virtual environment from it."
+        },
+        {
+            "name": "RezPackages",
+            "type": "STRING",
+            "userInterface": {
+                "control": "HIDDEN"
+            },
+            "default": "aftereffects-{{AE_VERSION}}",
+            "description": "If a queue accepts this parameter, it will create a Rez environment from it."
```
crowecawcaw: I think this should be a separate PR since it's not related to JSON vs YAML.

It'd also help to introduce this change alongside a README update describing how to use it. I can't quite imagine how it's supposed to work. The screenshot where both conda and rez packages are displayed as inputs at the same time doesn't seem right at first glance.

crowecawcaw (changes requested):

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.json
eherozhao: > The screenshot where both conda and rez packages are displayed as inputs at the same time doesn't seem right at first glance.

That is because I have both Conda Packages and Rez Pakcages in my QEnv during the test.

eherozhao: Thanks for your reply! I removed RezPackage change from this PR and will push a new PR related to this later.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1259,9 +1221,81 @@ function SubmitSelection(selection, framesPerTask) {
...
+            logger.debug("Error accessing the template's steps name. " + error, submitBundleFile);
+        }
+        const aftereffectsVersion = app.version[0] + app.version[1];
+        logger.debug("The major version of After Effects is " + aftereffectsVersion, submitBundleFile);
+
+        var paramDefCopy = templateDict.parameterDefinitions;
+        for (var i = paramDefCopy.length - 1; i >= 0; i--) {
+            if (paramDefCopy[i].name == "CondaPackages") {
+                paramDefCopy[i].default = "aftereffects=" + aftereffectsVersion;
+                alert(paramDefCopy[i].default);
```
mwiebe: Something left over from debugging?

eherozhao: Thanks for catching this, I have deleted it in the new revision.

src/submission/SubmitBundle.jsx
```diff
@@ -149,9 +111,76 @@ function SubmitSelection(selection, framesPerTask) {
...
+        try {
+            if (templateDict.steps[0].name) {
+                templateDict.steps[0].name = compName;
+                logger.debug("The step name is " + templateDict.steps[0].name, submitBundleFile);
+            }
+        } catch (e) {
+            adcAlert("Error accessing the template's steps name. \nPlease check your template.json and make sure you have name under steps.");
+            logger.debug("Error accessing the template's steps name. " + error, submitBundleFile);
+        }
+        const aftereffectsVersion = app.version[0] + app.version[1];
```
crowecawcaw, line 140: Do we need to add a `.` between the two version strings? What does the combined version look like?

src/submission/SubmitBundle.jsx
```diff
@@ -149,9 +111,76 @@ function SubmitSelection(selection, framesPerTask) {
...
+
+        var paramDefCopy = templateDict.parameterDefinitions;
+        for (var i = paramDefCopy.length - 1; i >= 0; i--) {
+            if (paramDefCopy[i].name == "CondaPackages") {
+                paramDefCopy[i].default = "aftereffects=" + aftereffectsVersion;
+            }
+        }
+
+        template.open("w");
+        template.write(JSON.stringify(templateDict));
```
crowecawcaw: Small nicety: the JSON will be nicely formatted if you do `JSON.stringify(templateDict, null, 2)`

src/submission/SubmitBundle.jsx
```diff
@@ -149,9 +111,76 @@ function SubmitSelection(selection, framesPerTask) {
...
+     * Generates job template json file
+     **/
+    function generateTemplate(bundlePath) {
+        // Write the template.json file
+        var template = new File(bundlePath + "/template.json");
+        template.open("r");
+        var templateContents = template.read();
+        template.close();
+        // Parse the template string to JSON dict.
+        var templateDict = JSON.parse(templateContents);
```
crowecawcaw: Super nit: JavaScript calls these "objects" instead of "dicts".

src/submission/SubmitBundle.jsx
crowecawcaw: Also do we need to add a `*` to the end of the version string or have some sort of flexible version requirement? It looks like the version will be pegged to a single release.

crowecawcaw (changes requested):

src/submission/SubmitBundle.jsx
eherozhao: Oh, goot to know! Will update this.

src/submission/SubmitBundle.jsx
eherozhao: Unfortunately we have to. The app.version[0] returned 2 and the next index returned 4 so the combination is version 24. It depends on if customer would like to use major version as the conda package version or also with a minor version.

I can keep this as a note and check if we would like to use major,minor version as the default conda package or just major version in the future. For now, I personally don't worry much about it.

src/submission/SubmitBundle.jsx
eherozhao: For sure. I can update this. Any thoughts on 4 spaces?

viknith (commented):
QQ, does setting RezPackages in QE and then seeing it autofill in the GUI submitter follow the pattern that other submitters follow? Asking because of the prior conversation we had about not wanting to repeat the same behavior with the CondaPackages parameter. @eherozhao @mwiebe

mwiebe (approved):

mwiebe:
> QQ, does setting RezPackages in QE and then seeing it autofill in the GUI submitter follow the pattern that other submitters follow? Asking because of the prior conversation we had about not wanting to repeat the same behavior with the CondaPackages parameter. @eherozhao @mwiebe

@eherozhao is planning to make a follow-up PR that makes the pattern for Rez identical to Conda.

crowecawcaw (approved):

PR #90, opened 2025-01-07
feat: Test assets for verifying aerender base functionality, plugin support, and custom font support.

cilevitz:
There are no standard example assets available for users to test After Effects render functionality on their farm.

We built a standard test project with several compositions, each demonstrating different types of functionality. For example, the plugin demonstration composition contains a plugin-based effect layer.

Users can verify correct render output when using plugins and custom fonts on their farm.

Testing was performed with SMF hosts running After Effects aerender to verify correct output.

No changes were made to the codebase.

Yes, there is a README.md describing the test project and its contents.

No.

----


cilevitz:
Closing per request.

PR #91, opened 2025-01-07
feat: Add custom font support by detecting fonts used in a composition and submitting them as job attachments

cilevitz:
After Effects requires all of a project's needed fonts to be installed system-wide onto the machine, or into the profile of the user running the software at render time. There is no font embedding available, nor does After Effects do any path searching for fonts.

[Pull request #46](https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/46) includes font files as job attachments in the bundle. We integrated it along with some fixes into the dockable submitter.

Users can use any custom fonts they'd like in their comp, and all chosen fonts will render correctly on the farm.

Testing was performed with SMF hosts running After Effects aerender to verify correct output of a project containing custom fonts. The farm render output was compared to local renders from a machine with those custom fonts installed, and the farm output is as expected.

Yes, there are changes to the project README.md describing the feature.

No.

----


dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,245 @@
...
+                fonts.add(full_assetpath)
+    
+    return fonts
+
+
+def install_font(src_path, scope=INSTALL_SCOPE_USER):
+    """
+    Install provided font to the worker machine
+
+    :param src_path: path of font that needs to be installed
```
viknith, line 117: Nit: missing :returns:

Same for some of the functions below

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        cb.value = ctypes.sizeof(is_truetype)
+        gdi32.GetFontResourceInfoW(
+            filename, ctypes.byref(cb), ctypes.byref(is_truetype), GFRI_ISTRUETYPE
+        )
+        if is_truetype:
+            fontname += " (TrueType)"
+
+        # Creates registry if it doesn't exist, opens when it does exist
+        with winreg.CreateKeyEx(registry_scope, FONTS_REG_PATH, 0, access= winreg.KEY_SET_VALUE) as key:
+            winreg.SetValueEx(key, fontname, 0, winreg.REG_SZ, filename)
```
viknith, line 155: I recall a prior conversation where we did not want to allow registry keys to be modifiable due to security risks. Has this concern been resolved? @eherozhao

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,69 @@ function findJobAttachments(rootComp) {
                     }
                 }
             }
+            if (layer instanceof TextLayer){
+                var font = layer.text.sourceText.value;
+                var fontLocation = font.fontLocation;
```
viknith: Naive question, I saw the popup handling we have in the previous if block for missing footage. In a similar vein, can there be case where there are missing fonts in the font location? Asking because I've sometimes loaded up AfterEffects and the application notifies me that some fonts are missing, so a similar warning might be warranted here before they end up submitting a job.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/template.json
```diff
@@ -108,6 +108,32 @@
...
             }
         }
+    },
+    {
+        "name": "Install Fonts to Worker",
+        "description": "Install Fonts to Worker",
+        "script": {
+            "actions": {
+                "onEnter": {
+                    "command": "powershell",
```
eherozhao: q: is there a reason we cannot use call `python font_manager.py --install` directly?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,69 @@ function findJobAttachments(rootComp) {
...
+                // If the font location has a font extension, use the actual file name
+                if(fontLocation.indexOf(".otf") !== -1 || fontLocation.indexOf(".ttf") !== -1) {
+                    var fontLocationSplit = fontLocation.split("\\");
+                    var fontName = fontLocationSplit[fontLocationSplit.length - 1];
+                }
+                // Else use the family name for the temp file that will be created 
+                else {
+                    if (dcUtil.getAEVersion() >= 24){
```
eherozhao: Could you help me understand why do we check layer instanceOf TextLayer before checking the version first?

If I understand right, line 159 font.font, this font is defined inside above if block so when the code runs this line, the font is not defined yet.

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,69 @@ function findJobAttachments(rootComp) {
...
+                        var fontName = font.font + ".otf";
```
eherozhao: Do we assume all fonts will be .otf extension? Will this cause a problem if the font is ttf or other extensions?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,69 @@ function findJobAttachments(rootComp) {
...
+                }
+                fontsInComp.push([fontName, fontLocation]);
+            }
         }
     }
+
+    if (fontsInComp.length > 0){
+        // formatting collected fonts
+        var fontReferences = generateFontReferences(fontsInComp);
+        for (var j = 0; j < fontReferences.length; j++){
```
eherozhao: nit: could we just use i instead of j since there is no i used before?

src/submission/JobTemplateHelper.jsx
eherozhao: q: which version of AE we have tested the font solution?

src/utils/Utils.jsx
```diff
@@ -720,7 +720,16 @@ function __generateUtil() {
...
+    function getAEVersion() {
+        /* Return After Effects version as int */
+        var version = parseInt(app.version[0] + app.version[1]);
+
+        return version;
+    }
+
+
     return {
+        "invertObject": invertObject,
```
eherozhao, line 689: This is duplicate to the next line.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    try:
+        # Determine font destination 
+        if scope == INSTALL_SCOPE_SYSTEM:
+            dst_dir = FONT_LOCATION_SYSTEM
+            registry_scope = winreg.HKEY_LOCAL_MACHINE
+        else:
+            # Check if the Fonts folder exists, create it if it doesn't
+            if not os.path.exists(FONT_LOCATION_USER):
+                print("Creating User Fonts folder: %s" % FONT_LOCATION_USER, flush=True)
```
eherozhao: I noticed you were using print f-string other places. Could we make them consistent?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        
+        dst_path = os.path.join(dst_dir, os.path.basename(src_path))
+
+        # Copy the font to the Windows Fonts folder
+        shutil.copy(src_path, dst_path)
+
+        # Load the font in the current session, remove font when loading fails
+        if not gdi32.AddFontResourceW(dst_path):
+            os.remove(dst_path)
+            raise WindowsError('AddFontResource failed to load "%s"' % src_path)
```
eherozhao: same comment to use fstring for consistency.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        # Notify running programs
+        user32.SendMessageTimeoutW(
+            HWND_BROADCAST, WM_FONTCHANGE, 0, 0, SMTO_ABORTIFHUNG, 1000, None
+        )
+
+        # Store the fontname/filename in the registry
+        filename = os.path.basename(dst_path)
+        fontname = os.path.splitext(filename)[0]
+
+        # Try to get the font's real name
```
eherozhao: should we have already had the real name from the submitter side? Could you help me understand this?

eherozhao:
Could we have the following env tested?
[ ] AE version < 24 in SMF within Windows client
[ ] AE version < 24 in CMF within macOS client
[ ] AE version >= 24 in SMF within Windows client
[ ] AE version >= 24 in SMF within macOs client
[ ] font with different extensions: otf, ttf or any other AE supported version.
[ ] font from Creative Cloud App and custom font installed in the local workstation.

If you cannot test any one in above list, please help us understand the reason. Thanks.

dist/DeadlineCloudSubmitter.jsx
viknith, line 1: Nit: I just pulled this PR and ran the build script `python jsxbundler.py --source src/OpenAESubmitter.jsx --destination dist/DeadlineCloudSubmitter.jsx`
and there was a minor space difference in this generated file on Line 728 and Line 731. Could you please re-run this python command and push up the updated format change?

viknith (changes requested):
**TL:DR** I tried submitting jobs using fonts from my local MacOS machine and fonts supported by Adobe and neither fonts were found and the job did not render with the fonts as expected. Some guidance on how to test this PR would be appreciated.

---

**Long explanation:**

I tried submitting an AE 25.1 job from my Mac laptop with a custom font called 8 Heavy from Adobe Creative Cloud found here: https://fonts.adobe.com/fonts/eight that I added directly from Creative Cloud.

Then I added that font to my assets and then submitted a job. Under Job Attachements, I see the following input file attached called `/private/var/folders/r_/m1d3rb6s2nb559nj4hd77ptm0000gq/T/TemporaryItems/tempFonts/Users/viknith/Library/Application Support/Adobe/CoreSync/plugins/livetype/.w/.52471.otf`

I submitted the job, and it succeed, but the rendered output did not use the font that I added to my job. Seems like this approach of adding a font is not covered.

Here were the logs from the font step:
```
2025/01/09 11:43:34-08:00  
2025/01/09 11:43:34-08:00 ==============================================
2025/01/09 11:43:34-08:00 --------- Entering Environment: Install Fonts to Worker
2025/01/09 11:43:34-08:00 ==============================================
2025/01/09 11:43:34-08:00 ----------------------------------------------
2025/01/09 11:43:34-08:00 Phase: Setup
2025/01/09 11:43:34-08:00 ----------------------------------------------
2025/01/09 11:43:34-08:00 ----------------------------------------------
2025/01/09 11:43:34-08:00 Phase: Running action
2025/01/09 11:43:34-08:00 ----------------------------------------------
2025/01/09 11:43:34-08:00 Running command C:\Windows\System32\WindowsPowerShell\v1.0\powershell.EXE -File C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/installfonts.ps1 C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9
2025/01/09 11:43:34-08:00 Command started as pid: 2380
2025/01/09 11:43:34-08:00 Output:
2025/01/09 11:43:34-08:00 Running python script at C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/font_manager.py
2025/01/09 11:43:35-08:00 Running font script job: install
2025/01/09 11:43:35-08:00 Warning: couldn't recursively find tempFonts in subfolder: assetroot-3751af5bd81011e6d21d
2025/01/09 11:43:35-08:00 No custom fonts found, continuing task...
2025/01/09 11:43:35-08:00 Process pid 2380 exited with code: 0 (unsigned) / 0x0 (hex)
No newer logs at this moment.
```

There was also log saying `2025/01/09 11:44:04-08:00 WARNING:After Effects warning: Project has missing fonts.`

I thought the problem might be due to the fact that I pulled in a custom font from Adobe, so then I decided to use a font that was locally available on my laptop called Amazon Ember Thin. The job attachment path was `/private/var/folders/r_/m1d3rb6s2nb559nj4hd77ptm0000gq/T/TemporaryItems/tempFonts/Library/Fonts/AmazonEmber_Th.ttf`, but it resulted in the same issue with the same error messages. 

Did I misconfigure something here? If so, where should the fonts be located so that I know that they will be pulled correctly? Some clearer notes on how to test this would be much appreciated!

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
eherozhao: Good question! This looks fine to me, since this will be user-specific registry instead of admin registry.

src/submission/JobTemplateHelper.jsx
cilevitz: The indenting on the `else` was resolved in commit 46d834e. We have tested the font solution in After Effects 24.0.3.2.

src/submission/JobTemplateHelper.jsx
cilevitz: Adding an extension ensures the font_manager install script can find and install the font. Adobe fonts have no extension. The aerender font engine handles fonts properly regardless of their file's extension in testing.

src/submission/JobTemplateHelper.jsx
cilevitz: Updated in commit 46d834e.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
cilevitz: Updated in commit 46d834e.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
cilevitz: Updated in commit 46d834e.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
cilevitz: As part of the dynamic font installation process, we might have change the name from what After Effects reports.

Windows appends `" (TrueType)"` to a font's plaintext name if the font file's `FontResourceInfo` has the `GFRI_ISTRUETYPE` bit set.

The TrueType tag is not part of the font's name in After Effects, and we can't access the TrueType description bit to determine what its name will be after Windows processes it. We therefore handle the renaming at install and uninstall time.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
cilevitz: The font_manager scripts modify the user's registry `HKEY_CURRENT_USER`, not the system registry `HKEY_LOCAL_MACHINE`.

src/submission/JobTemplateHelper.jsx
cilevitz: We added a popup showing the user if any missing fonts are detected during job attachment collection in commit 46d834e.

cilevitz:
There was a platform issue identified with how fonts are named by the After Effects scripting API. There is different behavior on Mac vs. Windows, where font names contain the path to the font on Mac.

We have adjusted how fonts are collected and named and the issue should be resolved in commit 46d834e. Could you please test again with the same setup to confirm?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,85 @@ function findJobAttachments(rootComp) {
                     }
                 }
             }
+            if (layer instanceof TextLayer) {
```
eherozhao: https://ae-scripting.docsforadobe.dev/general/project.html#project-usedfonts this is the new feature provided after ae 24.5. Could we consider supporting this?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,85 @@ function findJobAttachments(rootComp) {
...
+                    if (os.indexOf("mac") !== -1) {
+                        // Mac prefixes the full source path to the name. Use only the file name
+                        var fontPrefixSplit = text.font.split("/");
+                        var font = fontPrefixSplit[fontPrefixSplit.length - 1];
+                    } else {
+                        // Windows doesn't need file name adjustment
+                        var font = text.font;
+                    }
+                    
+                    // Adobe fonts have no font extensions. Adding an extension makes them installable by font_manager.py
```
eherozhao: Is adobe fonts here meaning the fonts added from Adobe Creative Cloud app?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,85 @@ function findJobAttachments(rootComp) {
                     }
                 }
             }
+            if (layer instanceof TextLayer) {
+                var text = layer.text.sourceText.value;
+                var fontLocation = text.fontLocation;
+
+                // Matches a period followed by one or more alphanumeric characters at the end
+                var extensionRegex = /\.[a-zA-Z]+$/;
```
eherozhao: instead of using regex, can we use lastIndexOf to check the extension? 

```        
var lastDotIndex = fontLocation.lastIndexOf('.');
if (lastDotIndex !== -1) {
    fileExtension = fontLocation.substring(lastDotIndex);
}
```

will this cover the general case?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,85 @@ function findJobAttachments(rootComp) {
...
+ * @param fontPaths an array containing the actual location of the font file and the name that should be given to the temp copy per font
+ * @return an array of the temp font paths that were created
+ **/
+function generateFontReferences(fontPaths) {
+    // Create a temp folder where all used fonts get gathered
+    var _tempFontsFolder = dcUtil.normPath(Folder.temp.fsName + '/' + "tempFonts");
+    var formattedFontsPaths = [];
+    var tempFontPath = new Folder(_tempFontsFolder);
+    if (!tempFontPath.exists) {
+        tempFontPath.create();
```
eherozhao, line 479: nit: https://extendscript.docsforadobe.dev/file-system-access/folder-object.html#folder-object-constructors will this create the folder automatically if it doesn't exist? Have you had a chance to try it out?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,85 @@ function findJobAttachments(rootComp) {
...
+    }
+
+    // Copy the font files to the temp folder
+    for (var i = 0; i < fontPaths.length; i++) {
+        var fontName = fontPaths[i][0];
+        var fontLocation = fontPaths[i][1];
+
```
eherozhao, line 486: In initial PR from Alex, there is a if check when the font is in system folder https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/46/files#diff-7823b70daa71d676921826901ed94cb91b761bf80ac000141139f51768830885R152, may I ask if we need to consider about this in this PR implementation?

src/submission/JobTemplateHelper.jsx
eherozhao: Also, could we move this code snippet into a separate method?

src/submission/JobTemplateHelper.jsx
```diff
@@ -144,12 +145,85 @@ function findJobAttachments(rootComp) {
...
+                    // Adobe fonts have no font extensions. Adding an extension makes them installable by font_manager.py
+                    font = font + ".otf";
+                }
+                fontsInComp.push([font, fontLocation]);
+            }
+        }
+    }
+
+    if (fontsInComp.length > 0) {
+        if (app.fonts.missingOrSubstitutedFonts != "") {
```
eherozhao, line 141: this is checking the missing fonts in the client side. Could you help me understand why we want to check this iin the client side instead of the worker side?

eherozhao:
> **TL:DR** I tried submitting jobs using fonts from my local machine and fonts supported by Adobe and neither fonts were found and the job did not render with the fonts as expected. Some guidance on how to test this PR would be appreciated.
> 
> **Long explanation:**
> 
> I tried submitting an AE 25.1 job from my Mac laptop with a custom font called 8 Heavy from Adobe Creative Cloud found here: https://fonts.adobe.com/fonts/eight that I added directly from Creative Cloud.
> 
> Then I added that font to my assets and then submitted a job. Under Job Attachements, I see the following input file attached called `/private/var/folders/r_/m1d3rb6s2nb559nj4hd77ptm0000gq/T/TemporaryItems/tempFonts/Users/viknith/Library/Application Support/Adobe/CoreSync/plugins/livetype/.w/.52471.otf`
> 
> I submitted the job, and it succeed, but the rendered output did not use the font that I added to my job. Seems like this approach of adding a font is not covered.
> 
> Here were the logs from the font step:
> 
> ```
> 2025/01/09 11:43:34-08:00  
> 2025/01/09 11:43:34-08:00 ==============================================
> 2025/01/09 11:43:34-08:00 --------- Entering Environment: Install Fonts to Worker
> 2025/01/09 11:43:34-08:00 ==============================================
> 2025/01/09 11:43:34-08:00 ----------------------------------------------
> 2025/01/09 11:43:34-08:00 Phase: Setup
> 2025/01/09 11:43:34-08:00 ----------------------------------------------
> 2025/01/09 11:43:34-08:00 ----------------------------------------------
> 2025/01/09 11:43:34-08:00 Phase: Running action
> 2025/01/09 11:43:34-08:00 ----------------------------------------------
> 2025/01/09 11:43:34-08:00 Running command C:\Windows\System32\WindowsPowerShell\v1.0\powershell.EXE -File C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/installfonts.ps1 C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9
> 2025/01/09 11:43:34-08:00 Command started as pid: 2380
> 2025/01/09 11:43:34-08:00 Output:
> 2025/01/09 11:43:34-08:00 Running python script at C:\ProgramData\Amazon\OpenJD\session-2d81d4ce86804bd6b571854e4e8dba72vm8fx4y9\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/font_manager.py
> 2025/01/09 11:43:35-08:00 Running font script job: install
> 2025/01/09 11:43:35-08:00 Warning: couldn't recursively find tempFonts in subfolder: assetroot-3751af5bd81011e6d21d
> 2025/01/09 11:43:35-08:00 No custom fonts found, continuing task...
> 2025/01/09 11:43:35-08:00 Process pid 2380 exited with code: 0 (unsigned) / 0x0 (hex)
> No newer logs at this moment.
> ```
> 
> There was also log saying `2025/01/09 11:44:04-08:00 WARNING:After Effects warning: Project has missing fonts.`
> 
> I thought the problem might be due to the fact that I pulled in a custom font from Adobe, so then I decided to use a font that was locally available on my laptop called Amazon Ember Thin. The job attachment path was `/private/var/folders/r_/m1d3rb6s2nb559nj4hd77ptm0000gq/T/TemporaryItems/tempFonts/Library/Fonts/AmazonEmber_Th.ttf`, but it resulted in the same issue with the same error messages.
> 
> Did I misconfigure something here? If so, where should the fonts be located so that I know that they will be pulled correctly? Some clearer notes on how to test this would be much appreciated!

@cilevitz may I ask if adding fonts from Creative Cloud has been resolved in the latest commit?

eherozhao:
> Could we have the following env tested? [ ] AE version < 24 in SMF within Windows client [ ] AE version < 24 in CMF within macOS client [ ] AE version >= 24 in SMF within Windows client [ ] AE version >= 24 in SMF within macOs client [ ] font with different extensions: otf, ttf or any other AE supported version. [ ] font from Creative Cloud App and custom font installed in the local workstation.
> 
> If you cannot test any one in above list, please help us understand the reason. Thanks.

@cilevitz could you please update the checklist on which items you have tested and which not? Thanks!

src/submission/JobTemplateHelper.jsx
eherozhao: I was having a poc on font in the past with the following code snippet for ae 24.6.
```
var usedList = app.project.usedFonts;
for (var i = 0; i < usedList.length; i++) {
    var font = usedList[i].font;
    var fontFullName = font.fullName;
    var fontLocation = font.location;
    var fileExtension = "";
    var lastDotIndex = fontLocation.lastIndexOf('.');
    if (lastDotIndex !== -1) {
        fileExtension = fontLocation.substring(lastDotIndex);
    }
    var fontName = fontFullName + fileExtension;
    fontLocations.push([fontName, fontLocation]);
}
```
and it worked well to identify the font on macOS. I didn't test it on Windows and older AE but would like bring this solution to you if you are not aware of since this is pretty simple in my opinion to get the font and its extension. In this way, we can always use the right extension for the font name instead of adding .otf like line 178 if I understand it correctly.

src/submission/JobTemplateHelper.jsx
viknith: I believe this is based on the comment where I made that request here: https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/91#discussion_r1905797519. And I believe the same check happens on the worker side.

Do you disagree with this approach?

viknith (changes requested):
I have submitted jobs using an Adobe Font, using a locally available system font for `.ttf and .otf`, and combinations of those fonts together and the render jobs passed. 

However, the `.ttc` extension fails with the following error message. Here are the error logs:
```
2025/01/14 09:14:27-08:00  
2025/01/14 09:14:27-08:00 ==============================================
2025/01/14 09:14:27-08:00 --------- Entering Environment: Install Fonts to Worker
2025/01/14 09:14:27-08:00 ==============================================
2025/01/14 09:14:27-08:00 ----------------------------------------------
2025/01/14 09:14:27-08:00 Phase: Setup
2025/01/14 09:14:27-08:00 ----------------------------------------------
2025/01/14 09:14:27-08:00 ----------------------------------------------
2025/01/14 09:14:27-08:00 Phase: Running action
2025/01/14 09:14:27-08:00 ----------------------------------------------
2025/01/14 09:14:27-08:00 Running command "C:\Program Files\Python311\python.EXE" C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/font_manager.py install C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_
2025/01/14 09:14:27-08:00 Command started as pid: 3624
2025/01/14 09:14:27-08:00 Output:
2025/01/14 09:14:27-08:00 INFO:Looking for fonts to install...
2025/01/14 09:14:27-08:00 INFO:Installing font: C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\tempFonts\AmazonEmber_Th.ttf
2025/01/14 09:14:27-08:00 INFO:Installing font: C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\tempFonts\Athelas.ttc
2025/01/14 09:14:27-08:00 INFO:Installing font: C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\tempFonts\.52471.otf
2025/01/14 09:14:27-08:00 INFO:Installing font: C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\tempFonts\Helvetica.ttc
2025/01/14 09:14:27-08:00 Traceback (most recent call last):
2025/01/14 09:14:27-08:00   File "C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/font_manager.py", line 262, in <module>
2025/01/14 09:14:27-08:00     _install_fonts(session_dir)
2025/01/14 09:14:27-08:00   File "C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/font_manager.py", line 220, in _install_fonts
2025/01/14 09:14:27-08:00     raise RuntimeError(f"Error installing font: {msg}")
2025/01/14 09:14:27-08:00 RuntimeError: Error installing font: Traceback (most recent call last):
2025/01/14 09:14:27-08:00   File "C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\DeadlineCloudAESubmission\scripts/font_manager.py", line 111, in install_font
2025/01/14 09:14:27-08:00     raise WindowsError(f'AddFontResource failed to load "{src_path}"')
2025/01/14 09:14:27-08:00 OSError: AddFontResource failed to load "C:\ProgramData\Amazon\OpenJD\session-ff91a2afdc5f4f3ab2e082be5b6e9b01eawh40l_\assetroot-3751af5bd81011e6d21d\private\var\folders\r_\m1d3rb6s2nb559nj4hd77ptm0000gq\T\TemporaryItems\tempFonts\Helvetica.ttc"
2025/01/14 09:14:27-08:00  
2025/01/14 09:14:27-08:00 Process pid 3624 exited with code: 1 (unsigned) / 0x1 (hex)
```
I haven't tested the other font files `.fnt`, `.fot`, and `.fon` yet. These seems to be Windows font files, please test them before requesting further testing on the PR.

src/submission/JobTemplateHelper.jsx
cilevitz: Yes, we're referring to the fonts added through Adobe Creative Cloud.

Ahuge:
Hey @viknith it looks like .ttc fonts aren't supported on Windows as they are a MacOS specific font format.
https://superuser.com/questions/114603/how-to-install-os-x-ttc-font-on-windows-error-ttc-does-not-appear-to-be-a

Is Adobe Fonts giving you .ttc font files? It looks like there is a way to extract a ttc to multiple ttf files from it if needed

viknith:
Hi @Ahuge! I added Helvetica to my AE text within AE, which was under the provided list of fonts, and picked it because it was an available font on my MacOS computer under `/System/Library/Fonts` and I noticed that it was a `.ttc` file. Based on this line in `font_manager.py`
```
FONT_EXTENSIONS = [".otf", ".ttf", ".ttc", ".fon", ".fnt", ".fot"]
```
I assumed `.ttc` would be supported, so I went for it. That's where the error came in.

viknith:
@Ahuge Additionally, it seems like `.ttc` is supported in Windows based on Microsoft's documentation page: https://learn.microsoft.com/en-us/typography/opentype/spec/otff

Ahuge:
> Hi @Ahuge! I added Helvetica to my AE text within AE, which was under the provided list of fonts, and picked it because it was an available font on my MacOS computer under `/System/Library/Fonts` and I noticed that it was a `.ttc` file. Based on this line in `font_manager.py`
> 
> ```
> FONT_EXTENSIONS = [".otf", ".ttf", ".ttc", ".fon", ".fnt", ".fot"]
> ```
> 
> I assumed `.ttc` would be supported, so I went for it. That's where the error came in.

Fair enough.

I think we should review the following:
  - Explicitly filter out system fonts from our submitter side font gathering. 
  - Review the font file types the Adobe Fonts provides fonts as. I would bet that they provide platform independent fonts such as OTF and TTF. I have yet to find a case where Adobe Fonts has given me a font of a different format. **Note I have not used Adobe Fonts on MacOS so folks with MacOS could help verify this for me**
  - I also saw some questions in this thread about using the new Adobe Font API. In my old PR I added support for both ways to get font references due to my specific client requiring After Effects 2023. (https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/77/files#diff-7823b70daa71d676921826901ed94cb91b761bf80ac000141139f51768830885R134-R146). I think that including both APIs would be best at this point.

Ahuge:
> @Ahuge Additionally, it seems like `.ttc` is supported in Windows based on Microsoft's documentation page: https://learn.microsoft.com/en-us/typography/opentype/spec/otff

Ah! Good find, I wonder if we have to install it using a different method because it is a collection of Fonts. 

I will do some research into that

Ahuge:
Following up @viknith 

It looks like `AddFontResourceW` should support .ttc
https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-addfontresourcew#parameters





| File Extension	| Meaning | 
| ------------------- | ---------- |
| .fon | Font resource file. |
.fnt | Raw bitmap font file. |
.ttf | Raw TrueType file. |
.ttc | East Asian Windows: TrueType font collection. |
.fot | TrueType resource file. |
.otf | PostScript OpenType font. |
.mmm | Multiple master Type1 font resource file. It must be used with .pfm and .pfb files. |
.pfb |Type 1 font bits file. It is used with a .pfm file. |
.pfm | Type 1 font metrics file. It is used with a .pfb file. |


Are you able to upload that `.ttc` font that you tried to install? I can try to install it locally and see what the issue may be

viknith:
@Ahuge For some reason, Github won't let me upload `.ttc` files, but I revisited the test I did and realized that when i select Helvetica, After Effects defaults it to Helvetica Light, and I can change this to any boldness of Helvetica I like.

Screenshot 2025-01-15 at 10 32 19 AM

However, when I submit the job, the job attachments brings in the `.ttc` file.
Screenshot 2025-01-15 at 10 33 11 AM

It could be worth looking into extracting a `.ttf` from a `.ttc` since a `.ttc` seems to be a collection of `.ttf`s. But as you said, it should be supported.

viknith:
@Ahuge Here's a place that could work: https://github.com/prchann/fonts/blob/main/Helvetica/Helvetica.ttc

Ahuge:
Hey @viknith 
The `.ttc` font seemed to install for me locally.

Are you able to check the SHA256 of your local file?

```
C:\Users\ahugh>certutil -hashfile C:\Users\ahugh\Downloads\Helvetica.ttc SHA256
SHA256 hash of C:\Users\ahugh\Downloads\Helvetica.ttc:
abe8b92d1fee217962f869e8d40759afad5ca8bb074d5b580323f50c7fdff8e4
```

viknith:
@Ahuge Hmmm, doesn't seem to match. I can push up my ttc file to a branch for you to download, but could you just try rendering with the file you already have?
```
viknith@7cf34df03377 Downloads % shasum -a 256 Helvetica.ttc 
25eceb458d4baf628ee0b6a135a9f8ac5ec7b2826646720834bd2bf00dfc825a  Helvetica.ttc
```

Ahuge:
Hey @viknith 

Ok so I did a bit of research. turns out my minimal font installer that I made locally here had a bug and it wasn't actually installing the ttc.

So confirming that .ttc file won't install on my local Windows 11 computer
image


I cannot find a way to install that TTC file on Windows.
It sort of appears that the Microsoft .TTC spec may be different than the MacOS .TTC spec?

I have yet to find a .TTC file that I can install on Windows however.




I'd like to loop this back to asking if anyone has an Adobe Font that is a .TTC?
If not, I think we should only support OTF and TTF at this point.


I have been reading the .TTC specifications and while theoretically we can extract a .TTC to multiple .TTF files, I think we should hold off on that until we have customer requests for it.

Just my 2c

viknith:
Based on this comment `Adobe fonts have no font extensions. Adding an extension makes them installable by font_manager.py`

We may not have worry have Adobe fonts are. We agree, let's hold off on `.ttc` and provide support later.

cilevitz:
Environment testing checklist results:

* AE version < 24 in SMF within Windows client
Cannot test as we cannot download versions of After Effects prior to 2024 from Creative Cloud.

* AE version < 24 in CMF within macOS client, AE version >= 24 in SMF within macOs client
We cannot test in the Mac environment.

* AE version >= 24 in SMF within Windows client
After Effects 24.0.3 and 25.1 have been used on Windows to test in SMF.

* font with different extensions: otf, ttf or any other AE supported version.
TTF, OTF, FON, and Adobe Fonts have been tested in SMF. We have found the .TTC font format is not properly supported by Windows, and that there seem to be no publicly available .TTC format fonts which work with Windows.

* font from Creative Cloud App and custom font installed in the local workstation.
Adobe Fonts and third-party TTF and OTF fonts have been tested on Windows. Adobe distributes their Adobe Fonts as OTF format on Windows.

cilevitz:
> I have yet to find a .TTC file that I can install on Windows however.
> ...
> I have been reading the .TTC specifications and while theoretically we can extract a .TTC to multiple .TTF files, I think we should hold off on that until we have customer requests for it.

There is a script to extract .TTF files from a .TTC: https://github.com/yhchen/ttc2ttf/

We have not found any .TTC files that install on Windows, and their constituent .TTF files also do not install on Windows when extracted.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1143,9 +1121,150 @@ function findJobAttachments(rootComp) {
...
+    var fontExtensions = [".otf", ".ttf"];
+
+    // Windows also supports .fon files
+    var os = $.os.toLowerCase();
+    if (os.indexOf("windows") !== -1) {
+        fontExtensions.push(".fon");
+    }
+
+    // Avoid instances where the filename has a dot but no extension
+    // Some Adobe Fonts files have a dot followed by 5 numbers as its name with no extension (e.g. ".52741")
```
eherozhao: are we confident with this be the only pattern Adobe Fonts have?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -34,7 +34,8 @@
 FONT_LOCATION_USER = os.path.join(os.environ.get("LocalAppData"), "Microsoft", "Windows", "Fonts")
 
 # Font extensions supported in gdi32.AddFontResourceW 
-FONT_EXTENSIONS = [".otf", ".ttf", ".ttc", ".fon", ".fnt", ".fot"]
+# OpenType fonts without an extension can also be installed (e.g. Adobe Fonts)
+FONT_EXTENSIONS = [".otf", ".ttf", ".fon", ""]
```
eherozhao, line 39: why do we have a empty string in the list?

eherozhao:
> * AE version >= 24 in SMF within Windows client
>   After Effects 24.0.3 and 25.1 have been used on Windows to test in SMF.
> *

Have you had a chance test it in CMF with AE >=24?
wanna clarify: were testing pointing to older commit or this latest one?

eherozhao:
I pulled the PR and tested locally in my macOS with AE 24.6. My scene has a text layer with American TypeWriter font type, the submitter failed to open GUI submitter with no error or warning. Please check the exception is caught properly for any unsupported fonts.

src/submission/JobTemplateHelper.jsx
```diff
@@ -146,10 +146,198 @@ function findJobAttachments(rootComp) {
...
+    var fontLocations = [];
+    // app.project.usedFonts was introduced in 24.5. Fall back to scanning text layers if version is older
+    if (dcUtil.getAEVersion() >= 24.5) {
+        var usedList = app.project.usedFonts;
+        for (var i = 0; i < usedList.length; i++) {
+            var font = usedList[i].font;
+            var fontFullName = font.fullName;
+            var fontLocation = font.location;
+            if (!fontLocation) {
+                adcAlert(
```
eherozhao, line 168: I hit an issue on pulling the font in Windows AE 24.6.4.
My font location locally is under path `D:\Users\haozhao\AppData\Roaming\Adobe\User Owned Fonts` with name `Amazon Ember Regular-1207.otf`. However, the submitter threw 
```
fontLocation for Amazon Ember is empty. This font won't be added to the job.
```
from here. Why is its location empty?

eherozhao: I also tried to comment out this if block so that the legacy method got used. However, it still failed with FontLocation was empty.

src/submission/JobTemplateHelper.jsx
cilevitz: `app.usedFonts` is used on supported versions of AE in commit 39260e0.

src/submission/JobTemplateHelper.jsx
cilevitz: Resolved in commit 39260e0.

src/submission/JobTemplateHelper.jsx
cilevitz: There is no way to create a folder automatically if it doesn't exist. The Folder constructor returns an object representing a folder at the given path regardless of whether it exists in the filesystem, so we must check and create a folder at that path.

src/submission/JobTemplateHelper.jsx
cilevitz: We tested fonts managed by the Windows font system and found they are supported well. Any font that was installed to a Windows system font directory by the Windows font system is supported by Windows on that machine.

dist/DeadlineCloudSubmitter.jsx
cilevitz: Comment updated in commit 2940887.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
cilevitz: Adobe Fonts have a numeric ID as a file name with no extension, e.g. "38011"

src/submission/JobTemplateHelper.jsx
cilevitz: Resolved in commit 2940887.

src/submission/JobTemplateHelper.jsx
```diff
@@ -183,11 +183,84 @@ function getFontsFromFile() {
...
+ * @return Font metadata object, or null if there was an error
+ **/
+function getFontPaths() {
+    var scriptPath = scriptFolder + "/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py";
+    var scriptFile = new File(scriptPath);
+    if (!scriptFile.exists) {
+        adcAlert(
+            "Error: Missing font script at " + scriptFile.fsName + "\n"
+            + "\n"
+            + "Please ensure that Deadline Cloud Monitor is installed correctly.",
```
eherozhao: why do we error DCM here?

cilevitz: If the font script is missing, the job attachment system will process the rest of the project but cloud fonts without resolved fontLocations will not be added.

We are anticipating that the submitter and its assets will be deployed by the DCM installer in production. Since the installer will be deploying the submitter, the font script should be at its expected location if the rest of the submitter is there and installed correctly.

Ahuge: I see what you're saying.

Perhaps we just need to reword it. Technically the DCM installer and the submitter installer are separate. Should we call it
`"Please ensure that Deadline Cloud Submitters are installed correctly"` instead of DCM?

eherozhao: Alex is right. Monitor is a separate application only for customers to review their jobs.

src/submission/JobTemplateHelper.jsx
```diff
@@ -131,11 +131,279 @@ function findJobAttachments(rootComp) {
...
+            true
+        );
+        return null;
+    }
+    
+    var pythonExecutable = "python";
+    var os = $.os.toLowerCase();
+    if (os.indexOf("mac") !== -1) {
+        pythonExecutable = "python3";
+    }
```
eherozhao, line 301: may I ask why we have mac with python3 here?

viknith (approved):

src/submission/JobTemplateHelper.jsx
viknith: As of the latest macOS versions, the newest Python version included is typically Python 3.12. If it was an older version though, it should be set to use python. There should be python version detection here

src/submission/JobTemplateHelper.jsx
```diff
@@ -131,11 +131,279 @@ function findJobAttachments(rootComp) {
...
+ **/
+function getFontPaths() {
+    var scriptPath = scriptFolder + "/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py";
+    var scriptFile = new File(scriptPath);
+    if (!scriptFile.exists) {
+        adcAlert(
+            "Error: Missing font script at " + scriptFile.fsName + "\n"
+            + "\n"
+            + "Please ensure that Deadline Cloud Monitor is installed correctly.",
+            true
```
eherozhao: What is this true here? The method only allowed one argument.

cilevitz: The [alert](https://extendscript.docsforadobe.dev/user-interface-tools/window-class.html#alert) function supports three arguments and only two were being sent previously in `adcAlert()`. We added the third argument `errorIcon` to `adcAlert()` which causes the alert dialog to display as an error dialog when set.

eherozhao:
I tested in Windows and it failed silently. I have to record the screen to see the error.
image

eherozhao:
> I tested in Windows and it failed silently. I have to record the screen to see the error. image

image
I think that python I mentioned above is the reason it failed.

src/submission/JobTemplateHelper.jsx
```diff
@@ -131,11 +131,273 @@ function findJobAttachments(rootComp) {
...
+    if (output["error"]) {
+        adcAlert(
+            output["error"],
+            true
+        );
+        return null;
+    }
+    return output;
+}
+
```
viknith: This is incorrect error handling. if python was missing here, then there would be nothing assigned to the output variable, and if nothing is assigned, the if conditional statement will throw a KeyError which will not get caught. There needs to be a helper function that checks which python executable name exists (python3 or python) and then run the execution.

eherozhao:
SonarQube check failed because the python file has duplicate code https://sonarcloud.io/summary/new_code?id=aws-deadline_deadline-cloud-for-after-effects&pullRequest=91 can you also refactor it?

src/submission/JobTemplateHelper.jsx
```diff
@@ -131,11 +131,273 @@ function findJobAttachments(rootComp) {
...
+ **/
+function getLocationForFont(fontPostScriptName) {
+    var fontPath = null;
+    try {
+        // Get user-installed fonts
+        var fontPaths = getFontPaths();
+        if (!fontPaths) {
+            return null;
+        }
+        for (path in fontPaths) {
```
eherozhao: I am seeing 
> Add the "let", "const" or "var" keyword to this declaration of "path" to make it explicit.

in SonarQube check for this syntax.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Collect font metadata from the given font file.
+    Returns a dictionary of metadata.
+    """
+
+    result = {}
+
+    try:
+        t = ttLib.TTFont(font_path)
+    except:
```
eherozhao: can we specify the except here? 
Here is the sonarqube high finding
> Specify an exception class to catch or reraise the exception

src/submission/JobTemplateHelper.jsx
cilevitz: Changed to "Deadline Cloud Submitter" in eaa27ab.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
cilevitz: Exception specified in commit 589add2.

src/submission/JobTemplateHelper.jsx
cilevitz: Added `var` in commit 589add2.

src/submission/JobTemplateHelper.jsx
cilevitz: Resolved in commit 9a18cfd.

src/submission/JobTemplateHelper.jsx
cilevitz: Added Python version detection in commit 589add2.

src/submission/JobTemplateHelper.jsx
```diff
@@ -131,11 +131,372 @@ function findJobAttachments(rootComp) {
...
+            }
+        } catch (e) {
+            logger.error(e.message, jobTemplateHelperFile);
+            logger.debug("Where command output: " + outputWhere, jobTemplateHelperFile);
+        }
+    }
+
+    if (!(pythonFound || python3Found)) {
+        logger.error("No Python found on the path", jobTemplateHelperFile);
+        errorMessage =
```
eherozhao, line 241: > Add the "let", "const" or "var" keyword to this declaration of "errorMessage" to make it explicit.

eherozhao: Please check all other errorMessage variables and make the update.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -0,0 +1,179 @@
...
+        if verbose:
+            print(traceback.format_exc())
+        return result
+
+    # Collect name metadata from the name table
+    names_table = t["name"].names
+    raw_table = {}
+    try:
+        raw_table = {i:str(names_table[i]) for i in range(0, len(names_table))}
+    except:
```
eherozhao, line 75: This and the following except: were not changed.
> Specify an exception class to catch or reraise the exception

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -0,0 +1,179 @@
...
+            print(f"search_root: {search_root}")
+
+        font_results = get_fonts(search_root, verbose=verbose)
+        fonts.update(font_results)
+
+    return fonts
+
+
+if __name__ == "__main__":
+    verbose = False
```
eherozhao, line 171: nit: this verbose was not usable out of thie python file actually if I understand correctly.

eherozhao (approved):

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
cilevitz: Setting `verbose = True` is for debugging at the command line. It's not intended for use with the submitter.

PR #93, opened 2025-01-10
feat: image chunking

eherozhao:
Before we have native OpenJD feature Chunking (see RFC for more details https://github.com/OpenJobDescription/openjd-specifications/pull/54/files), we would like to have this feature in After Effects submitter for customers to have their image sequences chucked to increase the total rendering time.

1. The submitter allows customers to pass in a param Frames per task (ChunkSize) and then depending on this number and the range of frames, there will be chunked tasks and assigned to workers.
2. Rewrote PS script to Python to be more portable.

1. Customer can use chunking in AE submitter.

Submitted AE job with video output and image output.

The image job will be split to multiple tasks based on chunk size. Example as follows. 
image

I tested the following sceanrios:
1. 0-2999 frames for video output.
2. 0-2999 frames with 1000 chunk size => 3 chunks or 3 tasks
3. 0-2999 frames with 1 chunk size => 3000 chunks or 3000 tasks
4. 100-299 frames with 300 chunk size => 1 chunk or 1 task
5. 100-299 frames with 299 chunk size => 2 chunks or 2 tasks, the second task only has 1 frame.

Yes.

N/A

---


dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
```diff
@@ -0,0 +1,159 @@
...
+            "type": "PATH",
+            "objectType": "FILE",
+            "dataFlow": "OUT",
+            "userInterface": {
+                "control": "HIDDEN",
+                "label": "Output File",
+                "groupLabel": "Frame Range"
+            },
+            "default": "~\\Desktop\\output",
+            "description": "The render output destination"
```
mwiebe: Should this be a directory instead of a file?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
```diff
@@ -0,0 +1,159 @@
...
+            "name": "OutputFile",
+            "type": "PATH",
+            "objectType": "FILE",
+            "dataFlow": "OUT",
+            "userInterface": {
+                "control": "HIDDEN",
+                "label": "Output File",
+                "groupLabel": "Frame Range"
+            },
+            "default": "~\\Desktop\\output",
```
mwiebe: Probably better to delete this default.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
```diff
@@ -0,0 +1,159 @@
...
+        },
+        {
+            "name": "Frames",
+            "type": "STRING",
+            "userInterface": {
+                "control": "LINE_EDIT",
+                "label": "Start Frame - End Frame",
+                "groupLabel": "Frame Range"
+            },
+            "default": "1-10"
```
mwiebe: Probably better to delete this default.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -0,0 +1,115 @@
...
+def main():
+    parser = argparse.ArgumentParser(description="After Effects Render Script")
+    parser.add_argument("project", type=str, help="Project file path")
+    parser.add_argument("rqindex", type=int, help="Render queue index")
+    parser.add_argument("outputpath", type=str, default="", help="Output path")
+    parser.add_argument(
+        "frames",
+        type=str,
+        default="0-100",
+        help="The range of frames, the format is start-frame-end-frame",
```
mwiebe: The dashes make the format `start-frame-end-frame` a bit hard to understand. Maybe change it to `startFrame-endFrame`.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    parser.add_argument("--chunk-size", type=int, help="The number of frames per task")
+    parser.add_argument("--index", type=int, help="The index of the current task")
```
mwiebe: This is the starting frame of the chunk, right?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -0,0 +1,115 @@
...
+    # Determine if second parameter is a Python script or a range
+    range_value = args.frames
+    range_list = range_value.split("-")
+    start_frame, end_frame = int(range_list[0]), int(range_list[1])
+
+    if args.chunk_size is not None and args.index is not None:
+        try:
+            frames_length = end_frame - start_frame + 1
+            # if there is only 1 frame in the chunk
+            if args.chunk_size == 1 or args.index == frames_length:
```
mwiebe: The `or args.index == frames_length` doesn't look right to me. It looks like the `else` code might work correctly in all cases, and we don't need this `if` at all?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -0,0 +1,115 @@
...
+            text=True,
+        )
+        for line in process.stdout:
+            print(line, end="", flush=True)  # Print stdout line by line
+
+            # Check for specific errors or warnings in stdout
+            if "WARNING:After Effects warning" in line:
+                print(f"After Effects Warning: {line.strip()}", file=sys.stderr)
+            if line.startswith("aerender ERROR"):
+                print(f"Aerender Error: {line.strip()}", file=sys.stderr)
```
mwiebe: What's the purpose of printing the line a second time in these cases?

mwiebe (changes requested):

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
eherozhao: for after effects, it is using the output file path in the aerender so we just read the output file from AE application and set the parameter here.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
eherozhao: The problem is it is a hidden param right now so we have to have a default if I understand it right.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
eherozhao: Yes, I can delete this one for sure.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
eherozhao: Oh yes, makes sense.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
eherozhao: yes, starting frame of the chunk, I can update the help to make it clear.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
eherozhao: for this one, It seems a redundant one. I will delete it.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
eherozhao: After looking into it, I think you are right. It should be end_frame instead of frames_length in this case and the else should cover this case well. I will test it out to make the corresponding change.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
mwiebe: The path is a pattern, not the actual filename, is that correct? I think a directory parameter with a separate pattern might be better, similar to what's in https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/job_bundles/blender_render/template.yaml#L30

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
mwiebe: Why not make it visible? Changing the output directory seems useful to me.

eherozhao: The output directory was set from the composition settings so we just want the submitter to read it out and set it to this job parameter.

eherozhao: I changed the OutputDir to be a editable parameter in the new revision.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/image_template.json
eherozhao: After the testing, aerender cmd only cares about the output settings in the composition so having a pattern changeable in the job parameter with no functional usage will confuse customer so in the new revision, I only make directory a editable param.

mwiebe (approved):

npmacl (approved):

PR #95, opened 2025-01-13
feat: Plugin support with workflow for Conda package build and render

cilevitz:
After Effects requires a project's referenced plugins to be installed onto the render machine in a location that requires elevated permissions on Windows to modify. After Effects has a version-dependent plugin path in the system registry, which also requires elevated permissions to modify. We need to install plugin files to a location that After Effects can access at render time in a service-managed fleet environment, where elevated permissions are not available.

We have created a workflow for producing Conda packages that dynamically install plugins into an SMF job's render environment. After Effects can then load a submitted project's referenced plugins at render time.

Users can add supported plugins to their render jobs so their effects will render correctly in an SMF environment.

The workflow was tested on CPU-only and GPU-accelerated service-managed fleets with RSMB 6.4.4 as the packaged plugin. We confirmed that following the workflow gets farm resources set up correctly and submitted jobs produce correct render output.

Yes, there is a README.md describing the workflow and its prerequisites.

No.

----


docs/plugin-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+3. In the submitter dialog, specify the following conda packages: `aftereffects rsmb`.
+
+4. Adjust any needed settings and click `Submit` to submit the job to the farm.
+
+5. The After Effects render job will run with the `RSMB` plugin added to the worker machine's render environment.
+
+### RSMB rendering note: CPU vs. GPU
+
+Be sure to match the `Use GPU` setting in RSMB to your farm's capabilities. If you try to render a GPU-accelerated layer on a CPU-only fleet, your render job may freeze or render a green placeholder instead of the expected elements. Disabling the `Use GPU` option on the layer's effect, or enabling GPU usage on the fleet will resolve this.
```
eherozhao, line 267: could you provide more info on how to disable the Use GPU option on the layer's effect and how to enable GPU usage on the fleet? At least as far as I know for the latter one, we need to have a GPU windows fleet associated with the queue and set the host requirement to be gpu specific right?

viknith (commented):
Unrelated, but on the main README of the package [here](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/e456d90621135e1c5dc3f247404d47169d259319/README.md#submitter-usage), under the line `Navigate to the Edit menu > Preferences > Scripting & Expressions and enable the following options:`, it is "After Effects > Settings > Scripting & Expressions" for Mac. Please revise the README to ensure that there are instructions for Mac when they are needed.

docs/plugin-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+This section describes how to set up a build job for a plugin package and build it into a specified S3 bucket. This enables render jobs with conda packages specified to be able to get these from the Conda channel on the S3 bucket.
+
+1. Clone the [deadline-cloud-samples](https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/conda_recipes/README.md) repository and copy the `rsmb-6-4-4` folder into the `conda_recipes` folder.
+
+2. Install `RSMB` through their download link at: https://revisionfx.com/products/rsmb/after-effects/
+
+3. The `RSMB` installer will add 3 `.aex` files to `C:\Program Files\Adobe\Common\Plug-ins\7.0\MediaCore\RSMB6AE`. Zip these up into `RSMB_6_4_4_aex.zip`.
+If you don't have an archive manager in Windows, you can select the plugin files in Explorer, right-click them, and use `Send To > Compressed (zipped) folder`.
+
+4. Generate the `sha256` hash for this file with the following command in a PowerShell terminal:
```
viknith: The instructions provided here imply that the customer has a Windows machine. We want to include guidance on how to handle this if they own a MacOS machine. Taking inspiration from here would help to set up Windows machine: https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/conda_recipes/aftereffects-25.0/README.md, maybe by just linking it and swapping out relevant commands and names.

viknith: Especially, would you mind providing the exact PowerShell command to run in order to zip up all three files up? That will ensure minimal mistakes are made.

viknith:
This conda recipe folder should be merged into the deadline-cloud-samples package instead of here, and we could just make a README update here that's pointing to the example plugin support for AfterEffects.

eherozhao:
> This conda recipe folder should be merged into the deadline-cloud-samples package instead of here, and we could just make a README update here that's pointing to the example plugin support for AfterEffects.

For now, it is ok for us to keep it as a drafted PR here. We can consider about merging this to sample later.

viknith:
I followed the instructions in your README to submit a AE job using RSMB plugin by doing the following:

- Assets from [here](https://github.com/redtoastproductions/deadline-cloud-for-after-effects/tree/feature_test_assets)
- After Effects 25.1 using RSMB 6.4.4 submitted from my MacOS machine
- Added RSMB conda recipe to my conda channel bucket, this was the demo version without licensing server set up
- Submitted the RSMB job from your assets using the [dockable submitter branch](https://github.com/aws-deadline/deadline-cloud-for-after-effects/tree/feature_dockable_submitter) without the GPU box checked to my Windows CPU fleet.
- Rendered successfully!

~~However, I'm uncertain whether the plugin is actually being pulled because there's no watermark and I don't have a license server configured. Something to look into.~~

~~Edit: Tested with the license server, worked great!~~ 
after further investigation, seems like it wasn't copying RSMB to the correct plugin directory, so this PR will need to be fixed.

docs/plugin-support/README.md
```diff
@@ -0,0 +1,224 @@
...
+     This can be resolved by switching queues through the GUI or changing the farm in the config file.
+     
+     The following command is available when the Deadline Cloud Monitor has been installed:
+     
+     ```
+     # Open GUI to adjust default farm and queue if necessary
+     deadline config gui
+     ```
+
+7. A build job should now be running on your farm. Once completed, you can view your package by navigating to your S3 bucket, with its default location:
```
viknith: Please include this link for setting up proxy server: https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/smf-byol.html

I think a lot of this stuff can be deduped by just including this link

docs/plugin-support/README.md
cilevitz: We have added more information on the Use GPU option, how it can be set on the layer, and how to enable GPU usage on the farm.

docs/plugin-support/README.md
cilevitz: We have updated the build workflow to include Mac directions, and a PowerShell zip command for Windows.

docs/plugin-support/README.md
cilevitz: We have included a link to the bring your own license server documentation.

eherozhao:
> Unrelated, but on the main README of the package [here](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/e456d90621135e1c5dc3f247404d47169d259319/README.md#submitter-usage), under the line `Navigate to the Edit menu > Preferences > Scripting & Expressions and enable the following options:`, it is "After Effects > Settings > Scripting & Expressions" for Mac. Please revise the README to ensure that there are instructions for Mac when they are needed.

You need to rely on dockable_submitter branch's README. https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/feature_dockable_submitter/README.md?plain=1#L66-L67

cilevitz:
> > This conda recipe folder should be merged into the deadline-cloud-samples package instead of here, and we could just make a README update here that's pointing to the example plugin support for AfterEffects.
> 
> For now, it is ok for us to keep it as a drafted PR here. We can consider about merging this to sample later.

Moved to `deadline-cloud-samples` as https://github.com/aws-deadline/deadline-cloud-samples/pull/79

cilevitz:
Moved to `deadline-cloud-samples` as https://github.com/aws-deadline/deadline-cloud-samples/pull/79

PR #96, opened 2025-01-17
feat: add xml file for submitter installer

eherozhao:
We need to add AE to installer.

Added install_builder xml.
1. The installer will ask customers if they want to install the script to 2024 or 2025.
1. Then if the default ScriptUI Panels folder exist, we auto fill the directory path in the installer.
1. If not, leave the directoryParameter empty for customer to input.
1. We added a warning notification when customers is using user profile

We will be able to add AE to the installer.

dev build successfully.

https://github.com/user-attachments/assets/21a56809-7e03-4943-9990-1d115c851b7b

N/A

N/A
----

### DIFF
--- a/install_builder/deadline-cloud-for-after-effects.xml
+++ b/install_builder/deadline-cloud-for-after-effects.xml
@@ -14,8 +14,7 @@
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionDirectory allowWildcards="1">
-                    <origin>
-                        components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets</origin>
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets</origin>
                 </distributionDirectory>
             </distributionFileList>
         </folder>
@@ -27,8 +26,7 @@
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionDirectory allowWildcards="1">
-                    <origin>
-                        components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets</origin>
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets</origin>
                 </distributionDirectory>
             </distributionFileList>
         </folder>
@@ -41,8 +39,7 @@
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionFile>
-                    <origin>
-                        components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
                 </distributionFile>
             </distributionFileList>
         </folder>
@@ -55,8 +52,7 @@
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionFile>
-                    <origin>
-                        components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
                 </distributionFile>
             </distributionFileList>
         </folder>
@@ -92,12 +88,10 @@
                 <setInstallerVariable>
                     <name>after_effects_script_ui_directory_2024</name>
                     <ruleEvaluationLogic>or</ruleEvaluationLogic>
-                    <value>C:\Program Files\Adobe\Adobe After Effects 2024\Support
-                        Files\Scripts\ScriptUI Panels</value>
+                    <value>C:\Program Files\Adobe\Adobe After Effects 2024\Support Files\Scripts\ScriptUI Panels</value>
                     <ruleList>
                         <fileExists>
-                            <path>C:\Program Files\Adobe\Adobe After Effects 2024\Support
-                                Files\Scripts\ScriptUI Panels</path>
+                            <path>C:\Program Files\Adobe\Adobe After Effects 2024\Support Files\Scripts\ScriptUI Panels</path>
                         </fileExists>
                     </ruleList>
                 </setInstallerVariable>
@@ -110,24 +104,20 @@
                 <setInstallerVariable>
                     <name>after_effects_script_ui_directory_2025</name>
                     <ruleEvaluationLogic>or</ruleEvaluationLogic>
-                    <value>C:\Program Files\Adobe\Adobe After Effects 2025\Support
-                        Files\Scripts\ScriptUI Panels</value>
+                    <value>C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels</value>
                     <ruleList>
                         <fileExists>
-                            <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support
-                                Files\Scripts\ScriptUI Panels</path>
+                            <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels</path>
                         </fileExists>
                     </ruleList>
                 </setInstallerVariable>
             </actionList>
             <conditionRuleList>
                 <fileExists>
-                    <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support
-                        Files\Scripts\ScriptUI Panels</path>
+                    <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels</path>
                 </fileExists>
                 <fileExists>
-                    <path>C:\Program Files\Adobe\Adobe After Effects 2024\Support
-                        Files\Scripts\ScriptUI Panels</path>
+                    <path>C:\Program Files\Adobe\Adobe After Effects 2024\Support Files\Scripts\ScriptUI Panels</path>
                 </fileExists>
             </conditionRuleList>
             <ruleList>
--- a/install_builder/deadline-cloud-for-after-effects.xml
+++ b/install_builder/deadline-cloud-for-after-effects.xml
@@ -9,13 +9,24 @@
         <folder>
             <description>After Effects Plug-in</description>
             <destination>${after_effects_script_ui_directory}</destination>
-            <name>aftereffectpluginsassets</name>
+            <name>aftereffectsplugin</name>
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionDirectory allowWildcards="1">
                     <origin>
-                        components/deadline-cloud-for-after-effects/dist/*</origin>
+                        components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets</origin>
                 </distributionDirectory>
+            </distributionFileList>
+        </folder>
+        <folder>
+            <description>After Effects Plug-in Script</description>
+            <destination>${after_effects_script_ui_directory}</destination>
+            <name>aftereffectspluginscript</name>
+            <platforms>all</platforms>
+            <distributionFileList>
+                <distributionFile>
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
+                </distributionFile>
             </distributionFileList>
         </folder>
     </folderList>
@@ -59,16 +70,14 @@
     <parameterList>
         <stringParameter name="deadline_cloud_for_after_effects_summary" ask="0" cliOptionShow="0">
             <value>Deadline Cloud for After Effects
-                - Register the plug-in with After Effects by moving the script to the After Effects
-                ScriptUI folder
+- Register the plug-in with After Effects by moving the script to the After Effects ScriptUI folder
             </value>
         </stringParameter>
         <directoryParameter>
             <name>after_effects_script_ui_directory</name>
             <description>After Effects Scripts Directory</description>
             <explanation>Path to scripts directory in the After Effects resources directory. For
-                easiest
-                installation, After Effects should be installed before installing Deadline Cloud for
+                easiest installation, After Effects should be installed before installing Deadline Cloud for
                 After Effects. The default ScriptUI directory looks like "C:\Program Files\Adobe\Adobe
                 After Effects 2024\Support Files\Scripts\ScriptUI Panels". </explanation>
             <allowEmptyValue>0</allowEmptyValue>
--- a/install_builder/deadline-cloud-for-after-effects.xml
+++ b/install_builder/deadline-cloud-for-after-effects.xml
@@ -7,20 +7,20 @@
     <show>1</show>
     <folderList>
         <folder>
-            <description>After Effects Plug-in assets</description>
-            <destination>${after_effects_scripts_ui_directory}</destination>
+            <description>After Effects Plug-in</description>
+            <destination>${after_effects_script_ui_directory}</destination>
             <name>aftereffectpluginsassets</name>
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionDirectory allowWildcards="1">
                     <origin>
-                        components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets/*</origin>
+                        components/deadline-cloud-for-after-effects/dist/*</origin>
                 </distributionDirectory>
             </distributionFileList>
         </folder>
-        <folder>
+        <!-- <folder>
             <description>After Effects Plug-in Script</description>
-            <destination>${after_effects_scripts_ui_directory}</destination>
+            <destination>${after_effects_script_ui_directory}</destination>
             <name>aftereffectspluginscript</name>
             <platforms>all</platforms>
             <distributionFileList>
@@ -29,7 +29,7 @@
                         components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
                 </distributionFile>
             </distributionFileList>
-        </folder>
+        </folder> -->
     </folderList>
     <initializationActionList>
         <if>
@@ -68,7 +68,7 @@
             </value>
         </stringParameter>
         <directoryParameter>
-            <name>after_effects_scripts_directory</name>
+            <name>after_effects_script_ui_directory</name>
             <description>After Effects Scripts Directory</description>
             <explanation>Path to scripts directory in the After Effects resources directory. For
                 easiest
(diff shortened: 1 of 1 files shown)


install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,133 @@
+<component>
+    <name>deadline_cloud_for_after_effects</name>
+    <description>Deadline Cloud for After Effects 2024/2025</description>
```
joel-wong-aws: nit: to be consistent with other components/easier to change in the future, should this be `Deadline Cloud for After Effects 2024-2025`?

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,133 @@
...
+            <conditionRuleEvaluationLogic>or</conditionRuleEvaluationLogic>
+            <actionList>
+                <componentSelection>
+                    <deselect></deselect>
+                    <select>AfterEffects2024</select>
+                </componentSelection>
+                <setInstallerVariable>
+                    <name>after_effects_script_ui_directory</name>
+                    <ruleEvaluationLogic>or</ruleEvaluationLogic>
+                    <value>C:\Program Files\Adobe\Adobe After Effects 2024\Support Files\Scripts\ScriptUI Panels</value>
```
joel-wong-aws: I'm not familiar with the Adobe After Effects file structure. Does it require Administrative permissions to install to this location? If so, user installs will not work. We had this issue with the Cinema 4D installer.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,133 @@
...
+    <show>1</show>
+    <folderList>
+        <folder>
+            <description>After Effects Plug-in</description>
+            <destination>${after_effects_script_ui_directory}</destination>
+            <name>aftereffectsplugin</name>
+            <platforms>all</platforms>
+            <distributionFileList>
+                <distributionDirectory allowWildcards="1">
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter_Assets</origin>
```
joel-wong-aws: Where does `DeadlineCloudSubmitter_Assets` come from?

joel-wong-aws:
> Both AE 2024 and AE 2025 exists in the %ProgramData% path, 2025 path has higher priority than 2024 and auto fill the path directory. 

I don't see any reference to `%ProgramData%` in this PR?

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,133 @@
...
+                    <ruleList>
+                        <fileExists>
+                            <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels</path>
+                        </fileExists>
+                    </ruleList>
+                </setInstallerVariable>
+            </actionList>
+            <conditionRuleList>
+                <fileExists>
+                    <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels</path>
```
joel-wong-aws, line 117: It seems like there is some duplication in this logic. Our current logic is basically:
```
if AE2024 or AE2025 file path exists:
   if AE2024 file path exists:
       set path to AE 2024
   if AE2025 file path exists:
       set path to AE 2025
```

This also duplicates the file path 3 times.

Is there any way we can do the equivalent of the following logic instead?
```
AE2024 path = `C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels`
AE2025 path = `C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels`
if AE2025 path exists:
   use AE 2025
else if AE 2024 path exists:
   use AE 2024
else:
  use empty
```
While this is a little more nested in installbuilder, it reduces duplication.

I'd also be okay with
```
AE2024 path = `C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels`
AE2025 path = `C:\Program Files\Adobe\Adobe After Effects 2025\Support Files\Scripts\ScriptUI Panels`
if AE2024 or AE2025 file path exists:
   if AE2024 file path exists:
       set path to AE 2024
   if AE2025 file path exists:
       set path to AE 2025 
```
since it avoids more nesting down the line when new versions of AE are released

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: I followed the pattern keyshot was using https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/install_builder/deadline-cloud-for-keyshot.xml#L3. I can update it. We can also update keyshot one.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: it does since it is under disk C. It can be installed by Run As Admin.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: This is a script folder we have https://github.com/aws-deadline/deadline-cloud-for-after-effects/tree/feature_dockable_submitter/dist here is the link for the artifacts we need for the submitter: 1 folder and 1 jsx script.

eherozhao:
> > Both AE 2024 and AE 2025 exists in the %ProgramData% path, 2025 path has higher priority than 2024 and auto fill the path directory.
> 
> I don't see any reference to `%ProgramData%` in this PR?

yea, I referred it to `C:\Program Files`, I will update the PR description to more accurate to avoid confusing.

install_builder/deadline-cloud-for-after-effects.xml
joel-wong-aws: Thanks!

Here's the KeyShot fix! https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/154

install_builder/deadline-cloud-for-after-effects.xml
joel-wong-aws: User installs (without Admin permissions) are a critical flow. If this requires Admin permissions, we should find an alternative way to add this script via a user install.

install_builder/deadline-cloud-for-after-effects.xml
joel-wong-aws: Alternatively, we could add these files to all versions of AE on the machine; we have a similar feature request for KeyShot: https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/136

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: Based on our conversation with customers, artists don't have the permissions like IT admin to install apps, so like our installer, IT admin install them for artists. We also have the instructions on Github to guide how to move files to the right folders. I think the problem is they will not have the permission to move the file in either way.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,133 @@
...
+            <name>aftereffectspluginscript</name>
+            <platforms>all</platforms>
+            <distributionFileList>
+                <distributionFile>
+                    <origin>components/deadline-cloud-for-after-effects/dist/DeadlineCloudSubmitter.jsx</origin>
+                </distributionFile>
+            </distributionFileList>
+        </folder>
+    </folderList>
+    <initializationActionList>
```
viknith, line 72: For people that are not familiar with XML, can you include some comments to explain the conditional logic statements going on here?

viknith (approved):

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: tbh, XML is self explanatory if you combine the install-builder syntax. If it is still hard for you to understand, you can throw it to the chatbot if you want.

eherozhao: for example: https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide/_actions.html
> Initialization Actions - <initializationActionList>: Executes when the installer has started, just before the parsing of the command line options.

MarinMaksutaj:
Were you able to test using AE to submit a job after successfully installing the submitter? If so, can you provide some screenshots?

eherozhao:
> Were you able to test using AE to submit a job after successfully installing the submitter? If so, can you provide some screenshots?

I think it is out of scope of this PR. AE installation just move folder and script file to the given folder path and based on my test, it succeeds.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: I updated the xml to add all versions of AE on the machine. Please check the attached video record as a reference.

MarinMaksutaj (approved):

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,197 @@
...
+            <actionList>
+                <!-- After Effects 2024 selection and script directory -->
+                <componentSelection>
+                    <deselect></deselect>
+                    <select>AfterEffects2024</select>
+                </componentSelection>
+                <setInstallerVariable>
+                    <name>after_effects_script_ui_directory_2024</name>
+                    <ruleEvaluationLogic>or</ruleEvaluationLogic>
+                    <value>C:\Program Files\Adobe\Adobe After Effects 2024\Support
```
lucaseck: Does this work on multiline like this? I've had many issues before with whitespace between tags and values being used.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,197 @@
...
+                    value="${all_components} deadline_cloud_for_after_effects" />
+            </actionList>
+            <elseActionList>
+                <setInstallerVariable name="component(deadline_cloud_for_after_effects).show"
+                    value="0" />
+            </elseActionList>
+        </if>
+
+        <if>
+            <conditionRuleEvaluationLogic>or</conditionRuleEvaluationLogic>
```
lucaseck: super nit: moving this down to where the action conditions are defined or the conditions up would help with readability.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,197 @@
...
+                <fileExists>
+                    <path>C:\Program Files\Adobe\Adobe After Effects 2025\Support
+                        Files\Scripts\ScriptUI Panels</path>
+                </fileExists>
+                <fileExists>
+                    <path>C:\Program Files\Adobe\Adobe After Effects 2024\Support
+                        Files\Scripts\ScriptUI Panels</path>
+                </fileExists>
+            </conditionRuleList>
+            <ruleList>
```
lucaseck: Is there a reason for the Windows check to be in `ruleList` instead of the above `conditionRuleList` where it would be all together?

lucaseck (commented):
An idea for you since I'm not sure if having the user have to select a scripts directory for both 2024 and 2025 even if they only have one version installed is the best user experience. With deadline-cloud-for-houdini previously we [added subcomponents](https://github.com/aws-deadline/deadline-cloud-for-houdini/commit/244bc7130999f38e5d408b0e845939411b09d2a4) for each release year and the user had to explicitly select which years to install for. I could see similar logic being useful here to only request the scripts directory if the subcomponent for that year is selected. Curious if you had any other ideas around having to provide for all versions even if not using 2024 and 2025?

install_builder/deadline-cloud-for-after-effects.xml
lucaseck: Are we pre-emptively warning users or stopping the install if they select AE without running as an elevated user? Or will it just get to the copying stage and fail with a write access error that isn't super clear?

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: You are right. I copy/paste the xml from 1 folder to another one before making the change and the format seems auto changed. 😔  I will update it.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: Yes, it will stop the installation if it doesn't have enough permission. 
image

Talked offline. I will try to make a update to give customers a warning when AE is selected with User Install.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: Sure. I will make a change on this.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: My understanding is due to the `<conditionRuleEvaluationLogic> or </conditionRuleEvaluationLogic>`, it will look for the `<conditionRuleList>`. If we move Windows check to be in `conditionRuleList`, it will cause the logic wrong unless we have something like `if (AE2024 or AE2025) and windows`, but I am not sure if installbuilder can have this complex logic.

eherozhao:
> An idea for you since I'm not sure if having the user have to select a scripts directory for both 2024 and 2025 even if they only have one version installed is the best user experience. With deadline-cloud-for-houdini previously we [added subcomponents](https://github.com/aws-deadline/deadline-cloud-for-houdini/commit/244bc7130999f38e5d408b0e845939411b09d2a4) for each release year and the user had to explicitly select which years to install for. I could see similar logic being useful here to only request the scripts directory if the subcomponent for that year is selected. Curious if you had any other ideas around having to provide for all versions even if not using 2024 and 2025?

This looks very interesting. I like this idea. I can investigate on how to make AE installation work as Houdini one.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: https://github.com/user-attachments/assets/b7dfc047-197e-483d-949f-4f70302db31b

Here is the latest local build I have to throw a warning if customer is going to install AE with no admin. This change need to be made in another package so I will not include it in this PR.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,197 @@
...
+                - Register the plug-in with After Effects by moving the script to the After Effects
+                ScriptUI folder.
+            </value>
+        </stringParameter>
+
+        <!-- Directory parameter for After Effects 2024 -->
+        <directoryParameter>
+            <name>after_effects_script_ui_directory_2024</name>
+            <description>After Effects 2024 Scripts Directory</description>
+            <explanation>Path to the scripts directory in After Effects 2024 resources. Default
```
joel-wong-aws: Instead of adding the default in the explanation, can we just use the `<default>` field?
https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide/parameter.html

This also makes it easier to understand from the CLI installer help text.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,197 @@
...
+                directory: "C:\Program Files\Adobe\Adobe After Effects 2024\Support
+                Files\Scripts\ScriptUI Panels".</explanation>
+            <allowEmptyValue>1</allowEmptyValue>
```
joel-wong-aws: Is there a reason we allow empty variables here? Also, is the `<mustBeWritable>` option useful here?

joel-wong-aws (changes requested):

install_builder/deadline-cloud-for-after-effects.xml
joel-wong-aws: This doesn't seem like the right behaviour. In the video, we are clicking "User Install", but then the script gets saved to the `C:\Program Files\Adobe\Adobe After Effects\...`, which is a system install (right?). So then should we always block After Effects installation unless "System Install" is selected?

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: I simplified the logic with one if condition block now.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: I have updated the installer xml so this should be resolved.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: If I understand this tag correctly, this is to fill the default to the directory parameter but for our case, we just want to notify customer the default one in the explanation. We have another logic to fill the default path if the path really exists.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: I updated the logic so this should be resolved.

install_builder/deadline-cloud-for-after-effects.xml
```diff
@@ -0,0 +1,154 @@
...
+            <actionList>
+                <setInstallerVariable name="after_effects_script_ui_directory_2025" value="" />
+            </actionList>
+        </if>
+    </initializationActionList>
+    <parameterList>
+        <stringParameter name="deadline_cloud_for_after_effects_summary" ask="0" cliOptionShow="0">
+            <value>Deadline Cloud for After Effects 2024-2025
+- Compatible with Adobe After Effects 2024-2025
+- Register the plug-in with After Effects by moving the script to the After Effects ScriptUI Panel directory.</value>
```
joel-wong-aws: `Register` -> `Registers`

install_builder/deadline-cloud-for-after-effects.xml
joel-wong-aws: What happens if you put the directory in default instead?

eherozhao: It should always fill the path with the default one we put. However, you can see we check the path exists or not before assigning it to the param.

install_builder/deadline-cloud-for-after-effects.xml
eherozhao: I was following other dccs on this sentence. I noticed maya, houdini and keyshot had "Register" https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/install_builder/deadline-cloud-for-maya.xml#L78 and c4d had "Register". I think we need to make it consistent. "Registers" makes sense to me. I will update the rest of them later.

joel-wong-aws (approved):

PR #98, opened 2025-01-22
fix: replace %20 with space for Windows os

eherozhao:
We noticed that the space in AE file name got changed to %20 in Windows os.

Replace it by using the util method from old submitter

Windows Submitter can identify space in file name.

Local test in Windows with space in file name and macOS machine with %20 in file name.

N/A
N/A
----

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -672,43 +672,6 @@ function __generateUtil() {
         return fileName;
     }
 
-    /**
-     * Checks for given frame list if duplicate frames are present.
-     * @param {string} frameList: List of frames given in the UI or entire frame range of the comp.
-     * Returns either array filled with duplicates, or if no duplicates have been found empty string.
-     */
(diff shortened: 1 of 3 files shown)


viknith (approved):

crowecawcaw (commented):
`%20` is used for spaces in URL encoding. Can you see if other special characters also are encoded? We may want to use `urllib.parse.unquote` instead.

eherozhao:
> `%20` is used for spaces in URL encoding. Can you see if other special characters also are encoded? We may want to use `urllib.parse.unquote` instead.

yes, I did some investigation but didn't find a way to do it in js es3 as easy as this python api. I will talk with our partner with more knowledge on the AE but based on the original submitter they built, this looked like the only one. The worst case is we keep tracking this until we know other special characters.

npmacl (approved):

crowecawcaw:
In JS, there's a function `decodeURI` which might be what we want: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI

PR #99, opened 2025-01-24
chore: update GitHub issue templates

joel-wong-aws:
Updating the Github issue templates to be consistent with other repos.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -7,10 +7,12 @@ body:
     attributes:
       value: |
         Thank you for taking the time to fill out this bug report!
+
         ⚠️ If the bug that you are reporting is a security-related issue or security vulnerability,
         then please do not create a report via this template. Instead please
         notify AWS/Amazon Security via our [vulnerability reporting page](http://aws.amazon.com/security/vulnerability-reporting/)
         or directly via email to [AWS Security](aws-security@amazon.com).
+
   - type: textarea
     id: description
     attributes:
@@ -52,9 +54,10 @@ body:
       description: Please provide information on the environment and software versions that you are using to reproduce the bug.
       value: |
         At minimum:
-        1. Operating system: (e.g. Windows Server 2022; macOS Sequoia 15.1.1; etc.)
-        2. Version of After Effects:
-        3: Version of this package:
+        1. Operating system (e.g. Windows Server 2022; macOS Sequoia 15.1.1; etc.)
+        2. Version of After Effects
+        3. Version of this package
+
         Please share other details about your environment that you think might be relevant to reproducing the bug.
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,7 +1,7 @@
 name: "\U0001F680 Feature Request"
 description: Request a new feature
 title: "Feature request: (short description of the feature)"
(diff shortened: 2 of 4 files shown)


eherozhao:
Thanks for raising this PR. We are going to merge dockable_submitter branch to mainline and it has already had these change. https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/feature_dockable_submitter/.github/ISSUE_TEMPLATE/bug.yml?plain=1

joel-wong-aws:
> Thanks for raising this PR. We are going to merge dockable_submitter branch to mainline and it has already had these change. https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/feature_dockable_submitter/.github/ISSUE_TEMPLATE/bug.yml?plain=1

Thank you! I have re-targeted this PR to `feature_dockable_submitter` and resolved the merge conflicts :)

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -52,9 +54,10 @@ body:
       description: Please provide information on the environment and software versions that you are using to reproduce the bug.
       value: |
         At minimum:
-        1. Operating system: (e.g. Windows Server 2022; macOS Sequoia 15.1.1; etc.)
```
eherozhao, line 55: AE doesn't work on Linux so we don't need this change.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -1,7 +1,7 @@
 name: "\U0001F680 Feature Request"
 description: Request a new feature
 title: "Feature request: (short description of the feature)"
-labels: ["feature", "needs triage"]
```
eherozhao, line 4: This doesn't look necessary to me. May I ask if you have any strong reasons for us to change it?

joel-wong-aws: There is no `feature` label in GitHub. The label for a "new feature or request" (which is added by default to every GitHub repo) is `enhancement`: https://github.com/aws-deadline/deadline-cloud-for-after-effects/labels

Since the label `feature` doesn't exist in the labels, it won't show up on new GitHub issues, and I think it's better to use the existing `enhancement` label than create a mostly overlapping `feature` label.

.github/ISSUE_TEMPLATE/bug.yml
joel-wong-aws: Thank you! Updated!

eherozhao (approved):

.github/ISSUE_TEMPLATE/maintenance.yml
```diff
@@ -1,7 +1,7 @@
 name: "🛠️ Maintenance"
 description: Some type of improvement
 title: "Maintenance: (short description of the issue)"
-labels: ["feature", "needs triage"]
+labels: ["needs triage"]
```
viknith, line 4: Why can't this include `enhancement`?

joel-wong-aws: Typically, issues cut using this issue template are some sort of build/testing/dependencies update.

`enhancement` means "New feature or request", which doesn't quite match those types of issues. I think the better default is to not include a label at all.

The other alternative is to create a new `maintenance` label, something that I think would make sense to do for all Deadline/OpenJD repos but would have to get alignment on with the other teams.

viknith: Makes sense!

viknith (approved):

PR #104, opened 2025-01-27
ci: Add version bump and release workflows for AE

moorec-aws:
AE submitter is soon to not be a python repository which means it no longer fits into our existing re-usable workflows.

Update the version bump and release workflows.

AE can be version bumped and release.

Tested the workflows in a developer github account

Not required, these are CI changes

No

No

----


.github/workflows/release_bump.yml
```diff
@@ -12,14 +12,73 @@ on:
...
+        run: |
+          if [[ "${{ inputs.force_version_bump }}" != "" ]]; then
+            BUMP_ARGS="$BUMP_ARGS --${{ inputs.force_version_bump }}"
+          fi
+
+          # Backup actual changelog to preserve its contents
+          touch CHANGELOG.md
+          cp CHANGELOG.md CHANGELOG.bak.md
+
+          pip install -r ./requirements-release.txt
```
eherozhao, line 51: I don't think we need this. There is no requirements-release file anymore.

.github/workflows/release_publish.yml
```diff
@@ -0,0 +1,88 @@
...
+          git merge --ff-only origin/mainline -v
+          git push origin release    
+
+      - name: PrepRelease
+        id: prep-release
+        run: |
+          COMMIT_TITLE=$(git show -s --format='%s' HEAD)
+          NEXT_SEMVER=$(python -c 'import sys, re; print(re.match(r"chore\(release\): ([0-9]+\.[0-9]+\.[0-9]+).*", sys.argv[1]).group(1))' "$COMMIT_TITLE")
+
+          # The format of the tag must match the pattern in pyproject.toml -> tool.semantic_release.tag_format
```
eherozhao: Is this working with AE since AE doesn't have pyproject.toml?

eherozhao (commented):
A general question: I saw you mentioned you tested it. Is there a way I can see the result of the test workflow?

.github/workflows/release_bump.yml
moorec-aws: We will need it for not since we are relying on python-semantic-release at the moment.

.github/workflows/release_publish.yml
moorec-aws: We will still rely on `python-semantic-release` to do versioning for now. I've moved the settings to `semantic.toml`.

PR #105, opened 2025-01-28
chore: format the code

eherozhao:
The last commit merged in forgot to generate the artifact jsx and format it.
1. Format the code
2. add a step on README about how to suppress the warning about 

N/A
N/A
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This submitter creates and submits [OpenJD](https://github.com/OpenJobDescriptio
 
 ## Compatibility
 
-1. After Effects 24 - 25,
+1. After Effects 2024-2025,
 1. Python 3.9 or higher; and
 1. Windows or macOS operating system.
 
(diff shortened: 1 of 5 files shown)


viknith (approved):

README.md
```diff
@@ -80,6 +79,7 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
 1. Open the Deadline Cloud Submitter Panel by clicking **Windows > DeadlineCloudSubmitter.jsx**.
 1. Select your composition from the list and click **Submit**. You can hit the **Refresh** button to refresh the list.
 1. (Optional: for image sequences output types) you can specify the number of frames per task so that the job created by the After Effects submitter will create the tasks based on the number and then Deadline Cloud will assign the tasks to available workers to delegate the load.
+1. If you see a warning popup window with "You are about to run the script contained in file", you can suppress the warning by following the instruction in the following message. Go to Preferences > Scripting & Expressions, and turn off the "Warn User When Executing Files" preferences, and restart After Effects.
```
npmacl, line 82: nit: "following the instruction in the following message" sounds a little awkward, maybe something like "you can suppress the warning with the following steps:". 

This will also prevent the warning if a not deadline file is executed either right? It might be good to mention that.

npmacl (approved):

PR #106, opened 2025-01-28
feat: new dockable AE submitter with custom font and image sequence chunking

### LINKED ISSUE #39: Bug: Submitting files outside of StorageProfile requires me to type "y" on CMD.exe prompt instead of a GUI popup
### Expected Behaviour

When I submit files outside of the storage profile on other submitters I get a popup telling me so and asking if I wish to proceed.

### Current Behaviour

It appears to hang and if I check the `ae-pipe` text file it shows the contents in `Code Snippet`.
I can go to the CMD window and type `y` and return and it will go through, however that is non-intuitive and feels like a bug.
It should popup in the AE GUI

### Reproduction Steps

Submit a file in AE outside of your storage profile. It seems to just happen.

### Code Snippet

```
Submitting to Queue: AlexHughesQueue
Job submission contains 1 input files totaling 105.56 KB.  All input files will be uploaded to S3 if they are not already present in the job attachments bucket.

Files were specified outside of the configured storage profile location(s).  Please confirm that you intend to submit a job that uses files from the following directories:

Under the directory 'D:\Footage\The Creator':
	1 input file
	1 output directory

To permanently remove this warning you must only use files located within a storage profile location.

Do you wish to proceed? [y/N]: 
```

### LINKED ISSUE #73: Feature request: Movie File Output

### LINKED ISSUE #74: Maintenance: PIP install adaptor
### Description

The current instructions and pip installs are outdated and don’t reflect the latest branch. There are several fixes in the main branch on GitHub that address these issues. As it stands, the current pip installer doesn't allow After Effects to run properly, but the updated main branch includes the necessary fixes to resolve this.

### Solution

Please update the PIP installer or the documentation to indicate that the latest branch from GitHub should be used.

eherozhao:

The current submitter has couple issues need to be addressed.

1. it requires the adaptor but the adaptor was not functional well.
1. the submitter is calling deadline cloud cli under the scene for all operations, which caused low performance.
1. the UI has many features but not all of them works as expected.

We rewrote the submitter to be dockable and added the following features based on the customer input:
1. Custom font support: now fonts will be auto detected as a part of the job attachment within your job bundle. 
2. Instead of calling deadline cli for all operations, we call deadline gui submitter to simplify the submission process.
3. Image sequence chunking. For image sequence output type, you can choose frames per task to separate whole frames to chunks as different tasks to accelerate the rendering.
4. We added install builder xml and will include AE as a part of Deadline Cloud Submitter installer so that Windows users can use the installer to install AE scripts.

Unblock more customers to use After Effects with Deadline Cloud.

We did through manual test in macOS and Windows for AE 24.6.4 and AE 25.1. Here are the test cases from @viknith 

Test Case | Descriptions
-- | --
Submit Button + is Dockable | AE submitter is dockable, can select from a list of compositions and click the submit button to create & submit job bundle via popup Deadline GUI submitter.
Video Rendering | Can submit basic AE video composition render jobs and all video files are successfully added as job attachments to the job bundle
Audio Rendering | Can submit basic AE audio composition render jobs and all audio files are successfully added as job attachments to the job bundle.
Custom Font Solution | Can submit AE jobs with text that uses custom fonts with supported font file extensions and the font files are submitted with the job as job attachments so that the rendered output uses the custom fonts
Image Chunking | Can submit AE job and specify Frames per task (ChunkSize) and then verify that depending on this number and the range of frames, there will be chunked tasks and assigned to workers.

We have updated all documents to reference this change.

It is. We are making a huge change to help our customers.

----


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -0,0 +1,2454 @@
...
+            if (pythonVersion >= 3) {
+                // Correct version of Python is installed, use current pythonExecutable
+            } else if ((pythonVersion < 3) && python3Found) {
+                pythonExecutable = "python3";
+            } else {
+                errorMessage =
+                    "Error: Python 3 is required but only Python 2 was found.\n" +
+                    "\n" +
+                    "Please ensure that Python 3 or higher is installed correctly.";
+                adcAlert(errorMessage, true);
```
viknith, line 1173: This alert is popping up on AE 25.1 on Windows even though I have python3 and python in my $PATH. It doesn't block customers from submitting jobs but it's an annoying popup that serves no purpose. Currently working on a fix, but this should not block this PR.

viknith (dismissed):

viknith (approved):

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,261 @@
...
+FONT_LOCATION_SYSTEM = os.path.join(os.environ.get("SystemRoot"), "Fonts")
+FONT_LOCATION_USER = os.path.join(os.environ.get("LocalAppData"), "Microsoft", "Windows", "Fonts")
+
+# Font extensions supported in gdi32.AddFontResourceW
+# OpenType fonts without an extension can also be installed (e.g. Adobe Fonts)
+FONT_EXTENSIONS = [".otf", ".ttf", ".fon", ""]
+
+logger = logging.getLogger(__name__)
+
+def find_fonts(session_dir):
```
npmacl, line 43: Is there not a way to determine what fonts are being used by the project, rather than just grabbing all fonts on the system?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,261 @@
...
+    Install provided font to the worker machine
+
+    :param src_path: path of font that needs to be installed
+
+    :returns: boolean that represents if the font was installed and a string with any traceback that was created
+    """
+    try:
+        # Determine font destination
+        if scope == INSTALL_SCOPE_SYSTEM:
+            dst_dir = FONT_LOCATION_SYSTEM
```
npmacl, line 124: It seems like on the submitting machine, the found can be stored in any 3 of:
```
    "%APPDATA%/Adobe/CoreSync/plugins/livetype",
    "%APPDATA%/Adobe/User Owned Fonts",
    "%LOCALAPPDATA%/Microsoft/Windows/Fonts"
```

but on the worker they will all be placed in `"%LOCALAPPDATA%/Microsoft/Windows/Fonts"`. Do any methods of using a font in after effects result in an absolute/relative path to the font file being saved in the project? If this is the case, I would expect that to be broken by this.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -0,0 +1,261 @@
...
+    Looks for all font files that were sent along with the job
+
+    :param session_dir: the root folder in which to look for files
+
+    :returns: a set with all found fonts
+    """
+    fonts = set()
+    for subfolder in os.listdir(session_dir):
+        # Only look in assetroot folders
+        if not subfolder.startswith("assetroot-"):
```
npmacl, line 54: I would consider using the path mapping file (path found by evaluating {{Session.PathMappingRulesFile}} in the template) produced by openJD rather than directory diving into the assetroot directory

https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#731-value-references

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -0,0 +1,179 @@
...
+                continue
+        except Exception:
+            if verbose:
+                print(traceback.format_exc())
+            continue
+
+        if verbose:
+            print(f"search_root: {search_root}")
+
+        font_results = get_fonts(search_root, verbose=verbose)
```
npmacl, line 164: is there not a way to get the files an aftereffects project uses (or ideally all files that it uses), rather than scraping the system for font files?

eherozhao: This python script is aiming for the edge case that font files are under C:\Users\AppData\ which cannot be found by Adobe.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
eherozhao: The reason they added the first two paths was there was a edge case that font under AppData\Adobe cannot be found by the Adobe font location api.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
eherozhao: > {{Session.PathMappingRulesFile}}

do you mean the passed in argument can be openjd param?

npmacl (approved):

PR #107, opened 2025-01-28
chore: add adcAlert errorIcon argument

eherozhao:

There is a sonarqube alert about the errorIcon was not passed in adcAlert method but used in the downstream method. Even without this change, the code works.

Add an argument in adcAlert method. 

Manual test

when it's true
image

when it's false
image

N/A
N/A
----

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -270,7 +270,6 @@ function __generateUtil() {
         return _assetsList;
     }
 
-
     function getDescription() {
         /**
          * Get description data from UI.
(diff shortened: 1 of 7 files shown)


npmacl (approved):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1574,7 +1575,7 @@ function SubmitSelection(selection, framesPerTask) {
                 logger.debug("The step name is " + templateObject.steps[0].name, submitBundleFile);
             }
         } catch (e) {
-            adcAlert("Error accessing the template's steps name. \nPlease check your template.json and make sure you have name under steps.");
+            adcAlert("Error accessing the template's steps name. \nPlease check your template.json and make sure you have name under steps.", false);
```
viknith: A lot of these adcAlerts include the word "error" in them, implying that they are errors. Why do we want the error icon to be false?

eherozhao: <div class="wy-table-responsive" style="box-sizing: border-box; margin-bottom: 24px; max-width: 100%; overflow: visible !important; color: rgb(64, 64, 64); font-family: Lato, proxima-nova, &quot;Helvetica Neue&quot;, Arial, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">
errorIcon | Optional. When true, the platform-standard alert icon is replaced by the platform-standard error icon in the dialog. Default is false.
-- | --


</div><br class="Apple-interchange-newline">

https://extendscript.docsforadobe.dev/user-interface-tools/window-class.html#alert

it doesn't mean its not error. It's just if using platform-standard error icon in the dialog. If it is false, it will use AE dialog style. If it is true, in windows, it will be windows style.

viknith: Spoke offline, if the error message is explicitly a message with "error" in the text, it should be true. If it's not really an error and just a warning or a notification, we can have it as false.

viknith (approved):

PR #108, opened 2025-01-28
fix: removing buggy Python CLI validation check and fixing python file execution

viknith:
On Windows AE 25.1, when adding the font GT America Condensed Bold, its a missing font, and the submitter will start to run the `get_user_font.py` file to search for the font file. However in the process, it would verify which python executable to use to run. However, this approach is buggy and doesn't actually run `python --version` correctly and even though python is available on my machine, I get an extra AE error message about it. And despite that, the job submission goes forward with no issue.

Screenshot 2025-01-28 at 2 00 44 PM

It seems like verifying on the python3 executable validation is causing part of the problem. However, since the minimum requirement for this submitter to work is to have Python 3.9 installed, which implies that Python 3.9 will be installed. For simplicity and to ensure customers don't have to deal with more popups than necessary, we should just remove this for now, and we should leave that type of validation to the actual Deadline GUI submitter.

Additionally, the python file get_user_fonts.py was not being executed correctly and seems like it also doesn't work. Red Toast is working on debugging why the file may not be working, for now this solution will make sure the file is at least being executed and Windows AE25 customers won't be barraged with extra AE popups saying file is missing.

Less popups for customers that could confuse them. The photo above which pops up twice will now only pop up once. 

Ran it in Windows AE25.1 and Mac AE25.1 and it works. This issue did not show up on AE24, so I haven't tested this PR on that yet but I will.

No

No

----


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -747,18 +747,18 @@ var _DC_LOGGER_DEFAULT_BACKUP_COUNT = 5
 function Logger(logFileName, logDirectoryPath, maxBytes, backupCount) {
     /**
      * Basic logger implementation with file rotation based on byte size.
-     *
+     *
```
eherozhao: this space is not needed.

viknith: I tried to remove it by running Format Document but it wouldn't go away

viknith: fixed it!

eherozhao (dismissed):
No harm for us to start from a simple version and then improve it with more customers input.

viknith:
Looks like a lot of our errors came from potentially not running SystemCalls correctly: https://ae-scripting.docsforadobe.dev/general/system/#systemcallsystem
Thanks to @npmacl  for the documentation catch and testing! We may need hold off on this PR.

viknith:
These commits are ensuring the AE popups will not occur, however it doesn't fix the fact that the get_user_fonts.py file doesn't seem to be working correctly with Windows AE25.1, but we're looking into a fix.

viknith:
DCO to pass, @eherozhao will fix the incorrectly signed commits.

PR #110, opened 2025-01-29
fix: fix the panel ui when allow scripts to write files and access ne…

eherozhao:
We noticed a bug during testing that when the allow scripts to write files and access the network was disabled, the script panel couldn't be loaded successfully.
The problem is thisObj is not existing when building the UI, we should change it to this.

Fix the problem.

Yes, manual testing enabled access scripts to write files and disabled.
When it is disabled, it should look like below.
image

N/A

N/A

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -2413,8 +2413,8 @@ if (isSecurityPrefSet()) {
 } else {
     //Print an error message and instructions for changing security preferences
     var submitterPanel =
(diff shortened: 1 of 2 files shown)


viknith (commented):
Could you include the README updates from this PR? https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/111/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

Mine probably won't be merged

viknith (approved):

npmacl (approved):

PR #111, opened 2025-01-29
fix: removing Python CLI validation check and fixing python file execution

viknith:
On Windows AE 25.1, when adding the font GT America Condensed Bold, its a missing font, and the submitter will start to run the `get_user_font.py` file to search for the font file. However in the process, it would verify which python executable to use to run. However, this approach is buggy and doesn't actually run `python --version` correctly and even though python is available on my machine, I get an extra AE error message about it. And despite that, the job submission goes forward with no issue.

Screenshot 2025-01-28 at 2 00 44 PM

It seems like current implementation of the python3 executable validation is causing one of the popups. However, since the minimum requirement for this submitter to work is to have Python 3.9 installed, which implies that Python 3.9 will be installed. For simplicity and to ensure customers don't have to deal with more popups than necessary, we should just remove this for now, and we should leave that type of validation to the actual Deadline GUI submitter. In the future though, we will want this validation to be added to make our code more robust.

Additionally, the python file get_user_fonts.py was not being executed correctly and seems like it also doesn't work. Red Toast is working on debugging why the file may not be working, for now this solution will make sure the file is at least being executed and Windows AE25 customers won't be barraged with unnecessary AE popups saying file is missing.

Less popups for customers that could confuse them. The photo above which pops up twice will now only pop up once. 

Ran it in Windows AE25.1 and Mac AE25.1 and it works. This issue did not show up on AE24, so I haven't tested this PR on that yet but I will.

No

No

----


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1211,7 +1116,15 @@ function getFontPaths() {
...
     try {
-        var outputRaw = system.callSystem(pythonExecutable + " \"" + scriptFile.fsName + "\"");
+        var os = $.os.toLowerCase();
+        var getUserFontsExecutionCommand;
+        if (os.indexOf("windows") !== -1) {
+            getUserFontsExecutionCommand = "start /min cmd.exe /c \"" + pythonExecutable + " \"" + scriptFile.fsName + "\"\"";
+        } else {
+            getUserFontsExecutionCommand = "/bin/sh -c '" + pythonExecutable + "\"" + scriptFile.fsName + "\"'";
+        }
+        adcAllert(getUserFontsExecutionCommand, false);
```
npmacl: Why are we calling adcAllert here? (also shouldnt it be adcAlert? why two ls?)

src/submission/JobTemplateHelper.jsx
```diff
@@ -304,7 +209,15 @@ function getFontPaths() {
...
-        var outputRaw = system.callSystem(pythonExecutable + " \"" + scriptFile.fsName + "\"");
+        var os = $.os.toLowerCase();
+        var getUserFontsExecutionCommand;
+        if (os.indexOf("windows") !== -1) {
+            getUserFontsExecutionCommand = "start /min cmd.exe /c \"" + pythonExecutable + " \"" + scriptFile.fsName + "\"\"";
+        } else {
+            getUserFontsExecutionCommand = "/bin/sh -c '" + pythonExecutable + "\"" + scriptFile.fsName + "\"'";
+        }
+        adcAllert(getUserFontsExecutionCommand, false);
+        var outputRaw = system.callSystem(getUserFontsExecutionCommand);
```
npmacl: Since this pattern appears more than once, I recommend writing a function for it. Then using it here would look something like:
```
var outputRaw = runShellCmd(pythonExecutable + "\"" + scriptFile.fsName + "\"'");
```

src/submission/JobTemplateHelper.jsx
```diff
@@ -304,7 +209,15 @@ function getFontPaths() {
...
     try {
-        var outputRaw = system.callSystem(pythonExecutable + " \"" + scriptFile.fsName + "\"");
+        var os = $.os.toLowerCase();
+        var getUserFontsExecutionCommand;
+        if (os.indexOf("windows") !== -1) {
+            getUserFontsExecutionCommand = "start /min cmd.exe /c \"" + pythonExecutable + " \"" + scriptFile.fsName + "\"\"";
+        } else {
+            getUserFontsExecutionCommand = "/bin/sh -c '" + pythonExecutable + "\"" + scriptFile.fsName + "\"'";
+        }
+        adcAllert(getUserFontsExecutionCommand, false);
```
npmacl: Similar question about why we have adcAllert here.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1091,118 +1091,23 @@ function getFontsFromFile() {
...
  **/
 function getFontPaths() {
     var errorMessage = "";
-    // Ensure Python exists and is at least version 3
-    var pythonExecutable = getPythonExecutable();
-    if (!pythonExecutable) {
+    // Python 3.9 is the minimum requirement for the Deadline GUI submitter,
+    // Eventually, we want better validation by check whether the customer has python3 or
+    // python in their $PATH.
+    var pythonExecutable = "python3";
```
npmacl, line 1115: I think it would be better to accommodate if python is not installed under the name python3. For example on my windows laptop I have installed the deadline cloud CLI, the DCM, and I have python on my path and it is version 3.9+:
```
C:\Users\npmac>python -V
Python 3.10.8
```

I also have a file python3 on my path as well (path `C:\Users\npmac\AppData\Local\Microsoft\WindowsApps\python3.exe)`, but this seems to just be an executable that opens the windows store to install python there. I think this is something Microsoft sets up by default.

src/submission/JobTemplateHelper.jsx
viknith: Good idea!

src/submission/JobTemplateHelper.jsx
viknith: The file in the dist folder gets generated from the code here

dist/DeadlineCloudSubmitter.jsx
viknith: Agreed! I had been hitting issues with trying to make Python validation work since there are various edge cases to concern, such as executable verification, version verification, and also ensuring that under "Manage app exection aliases" the python3 and python ones are disabled. I'll cut a ticket for that in our backlog to fix this, but for now it's lower priority.

README.md
```diff
@@ -65,12 +65,16 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
...
 
+   Additionally, to disable warnings every time you submit a job with the submitter, deselect the following: `Select Edit > Preferences > Scripting & Expressions > deselect Warn User When Executing Files`
+
 1. Copy `DeadlineCloudSubmitter.jsx` and the `DeadlineCloudSubmitter_Assets` folder in the `dist` folder to
    the **ScriptUI Panels** folder within your After Effects installation. This folder is typically located at the following path:
 
    - Windows: `Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`
    - macOS: `Applications/Adobe After Effects <version>/Scripts/Script UI Panels`
 
+1. Run `pip install fonttools` in your local Terminal or Command Prompt to install the fonttools library  used by the submitter to handle finding fonts.
```
eherozhao: nit: there's a extra space between library and used. We can also call out that this only needed if you see the prompt warning window.

dist/DeadlineCloudSubmitter.jsx
eherozhao: might be debug usage?

viknith: Mistake, will fix!

README.md
viknith: But we don't want customers to ever hit this warning window, so it makes more sense to just tell them to install now rather than later so that they don't have to deal with troubleshooting.

src/utils/Utils.jsx
```diff
@@ -1,5 +1,17 @@
...
+    var os = $.os.toLowerCase();
+    var finalCommand;
+    if (os.indexOf("windows") !== -1) {
+        finalCommand = "start /min cmd.exe /c \"" + command + "\"";
+    } else {
+        finalCommand = "/bin/sh -c '" + command + "'";
+    }
+
+    return system.callSystem(finalCommand);
+}
```
npmacl, line 13: Why is this function defined twice?

viknith: The DeadlineCloudSubmitter.jsx file is autogenerated from the files under the src directory.

viknith:
Holding off on merging, this PR may be causing issues instead of resolving them.

viknith:
Closing because this PR doesn't work and we haven't been able to verify this issue on multiple devices.

PR #114, opened 2025-01-31
feat: add pyproject.toml to unblock installer build in the pipeline

eherozhao:

AE was not a python project which was not fitting into the current installer build pipeline.

Added the pyproject.toml and hatch.toml with other scripts to make sure the installer can be bulit in the pipeline.

AE installer can be built successfully in the pipeline.
1. local dev build.
2. local deployment the pipeline, point the source code to test-installer branch and trigger the pipeline. The installer was able to be built. Tested the installer in windows workstation and it can install AE submitter successfully.

N/A
N/A

---


depsBundle.py
```diff
@@ -0,0 +1,154 @@
...
+import shutil
+import subprocess
+import sys
+
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import Any
+
+SUPPORTED_PYTHON_VERSIONS = ["3.9", "3.10", "3.11", "3.12"]
+SUPPORTED_PLATFORMS = ["win_amd64", "manylinux2014_x86_64", "macosx_10_9_x86_64"]
```
viknith, line 13: Should this include linux? Same question for all other references to Linux.

moorec-aws (approved):

viknith (approved):

PR #116, opened 2025-02-01
fix: added UTF-8 file encoding support on all files written to or read from

### LINKED ISSUE #113: Bug: Special Characters cause issues/
### Describe the bug

Files with special character ( e.g. with "ñ" in the name) cause errors on submission.

### Expected Behaviour

I believe these character sets ( utf-8) should be supported.

### Current Behaviour

Error.

### Reproduction Steps

Submit a scene with a special character in the name.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; macOS Sequoia 15.1.1; etc.)
2. Version of After Effects
3. Version of this package

viknith:
Files with special characters were not being processed correctly. See #113 for more info.

Ensure that every file being read from and/or written to that is submitted or used for job bundle are set to UTF-8 encoding. Resolves #113 

Characters like ñ are now supported and won't break the AE job submission.

MacOS AE25 and Windows AE25 job submission, verified that the job template does pop up with the correct job attachment paths and the resulting job bundle JSON files do not have the missing character emoji.
Screenshot 2025-01-31 at 3 22 35 PM
Screenshot 2025-01-31 at 3 22 14 PM

No need

- Did you update all relevant docstrings for functions that you modified?
No need

- Should the README.md, DEVELOPMENT.md, or other documents in the repository's docs/ directory be updated along with your change?

No need

No

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -4,6 +4,24 @@
 
 var scriptFolder = Folder.current.fsName;
 
+function readFile(filePath) {
+    var f = new File(filePath);
+    f.encoding = "UTF-8";
+    f.open("r");
+    var fileContents = f.read();
+    f.close();
+    return fileContents;
+}
+
+function writeFile(filePath, fileContents) {
+    var f = new File(filePath);
+    f.encoding = "UTF-8";
+    f.open("w");
+    f.write(fileContents);
+    f.close();
+    return;
+}
+
 function timeToFrames(time, fps) {
     //temporarily change display format so we can convert seconds to frames
     //We could perform the math ourselves, but using After Effects's internal methods ensure that we don't lose precision due to floating point errors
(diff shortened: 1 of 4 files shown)


crowecawcaw (dismissed):

eng-josh (approved):

NathanMacLeod (approved):

npmacl (approved):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1532,7 +1540,7 @@ function SubmitSelection(selection, framesPerTask) {
             sanitizedOutputFolder
         );
         var assetReferencesOutDir = bundlePath + "/asset_references.json";
-        writeJSONFile(jobAttachmentsContents, assetReferencesOutDir);
+        writeFile(assetReferencesOutDir, JSON.stringify(jobAttachmentsContents));
```
eherozhao, line 1543: I think we deleted two argument (null, 4). What are the intentation of the new files?

viknith: Ah I missed this, let me fix it.

viknith: null, 4 just makes the files prettier to look at. Without it, it would look uglier. I'll put up a chore PR to fix it.

PR #117, opened 2025-02-01
chore: add JSON stringify spacing parameter to make prettier JSON files

viknith:
In previous PR [here](https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/116), I forgot to add some parameters to JSON.stringify. Results in uglier JSON files.

Added them

Cleaner JSON files

Not tested

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1540,7 +1540,7 @@ function SubmitSelection(selection, framesPerTask) {
             sanitizedOutputFolder
         );
         var assetReferencesOutDir = bundlePath + "/asset_references.json";
-        writeFile(assetReferencesOutDir, JSON.stringify(jobAttachmentsContents));
+        writeFile(assetReferencesOutDir, JSON.stringify(jobAttachmentsContents, null, 4));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


viknith:
I compared the JSON files before and after and there's virtually no difference. Verified on Windows and Mac.

This is before:
Screenshot 2025-01-31 at 4 42 35 PM

This is after:
Screenshot 2025-01-31 at 4 47 45 PM

npmacl (approved):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -2444,7 +2450,7 @@ if (isSecurityPrefSet()) {
     errorText2.text = [
         "In order for the Deadline Cloud submitter to execute, you need to update your script permissions to allow script networking and file access. To do this, follow the instructions below",
         "  1)  For Windows User: Select Edit > Preferences > Scripting & Expressions > select Allow Scripts To Write Files And Access Network",
-        "        For macOS User: Select After Effects > Settings > Scripting & Expressions > select Allow Scripts To Write Files And Access Network",
+        "       For macOS User: Select After Effects > Settings > Scripting & Expressions > select Allow Scripts To Write Files And Access Network",
```
eherozhao, line 2453: I intented leave one more space since it would align better on the AE side. Not a blocker. I will add it back with a comment.

eherozhao (approved):

dist/DeadlineCloudSubmitter.jsx
viknith: Unfortunately this will undo itself every time I format on VSCode. How would I prevent that?

PR #118, opened 2025-02-01
fix: Don't attach missing or substituted fonts on submit

cilevitz:
The submitter was collecting all font files in a project, including substituted fonts resulting from After Effects loading a project with missing or previous substituted fonts. Render tasks proceeded using the substituted fonts, producing incorrect output despite the tasks being marked as Succeeded.

We filter out any missing or substituted fonts from the list of used font files that gets processed when collecting job attachments. We also adjusted `call_aerender.py` so that it errors out if it detects the warning `Project has missing fonts` in the render log.

Users can use any custom fonts and cloud fonts they'd like, and all chosen fonts will render correctly on the farm. The user is alerted to any missing or substituted fonts before opening the submitter. Missing fonts will cause a render task to stop and be marked "Failed".

Testing was performed with SMF hosts running After Effects 24.0 and 25.1 to verify correct `aerender` output of a project containing custom and cloud fonts. The farm render output was compared to local renders from a machine with those fonts installed, and the farm output is as expected. The log output parser in `call_aerender.py` stops and fails the render task if a missing font warning is seen.

The README.md was updated with notes about missing and substituted fonts.

No.

----


dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -91,6 +91,11 @@ def main():
             if "WARNING:After Effects warning" in line:
                 print(f"After Effects Warning: {line.strip()}", file=sys.stderr)
 
+                # Some warnings should be treated as errors, i.e. a fatal warning
+                if "Project has missing fonts" in line:
```
eherozhao: Do we need to make sure in the submitter side, we have enough warning to customers that the submitted project with missing fonts will fail?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -72,7 +72,7 @@ def get_font(font_path):
     raw_table = {}
     try:
         raw_table = {i:str(names_table[i]) for i in range(0, len(names_table))}
-    except Exception:
+    except Exception as e:
```
eherozhao, line 85: q: if we don't print out error in the block, do we still need `as e`? I would suggest us either print the error out or delete `as e`.

Also, are we able to print out the python results from subprocess to the main submitter output to surface to customers?

src/submission/JobTemplateHelper.jsx
```diff
@@ -141,10 +141,24 @@ function findJobAttachments(rootComp) {
         if (app.fonts.missingOrSubstitutedFonts != "") {
             adcAlert("Missing fonts in project: " + (app.fonts.missingOrSubstitutedFonts).toString());
         }
+        // Remove missing or substituted fonts from fontsInProject to prevent incorrect render output.
+        // Build a new array containing only attachable fonts.
+        var fontsInProjectFiltered = [];
+        for (var i = 0; i < fontsInProject.length; i++) {
+            var attachFont = true;
+            // If the font's fontPostScriptName contains a missing or substituted font name, don't attach it
+            if (app.fonts.missingOrSubstitutedFonts.toString().indexOf(fontsInProject[i].fontPostScriptName) !== -1) {
```
viknith: Wouldn't the alternative behavior be to just name the files correctly with the fonts that will actually be used in place of the missing fonts? IIRC, that was the confusing part, since the substituted font files were being renamed to fonts that actually won't be included. It may be possible that customers are aware that their composition is missing the correct fonts for now, but don't want to modify the composition and allow Adobe to substitute in some default fonts for now and just get a job rendering for now. 

This behavior could be toggle-able in case customers still want to submit the job with substituted fonts and want them to pass, and at the same time we can avoid misnaming files. Might be out of scope of this PR, bu it is worth discussing.

src/submission/JobTemplateHelper.jsx
```diff
@@ -141,10 +141,24 @@ function findJobAttachments(rootComp) {
         if (app.fonts.missingOrSubstitutedFonts != "") {
             adcAlert("Missing fonts in project: " + (app.fonts.missingOrSubstitutedFonts).toString());
         }
+        // Remove missing or substituted fonts from fontsInProject to prevent incorrect render output.
+        // Build a new array containing only attachable fonts.
+        var fontsInProjectFiltered = [];
+        for (var i = 0; i < fontsInProject.length; i++) {
+            var attachFont = true;
```
viknith: Also this code can be simplified to an if/else structure, no need for extra attachFont boolean:
```
if (app.fonts.missingOrSubstitutedFonts.toString().indexOf(fontsInProject[i].fontPostScriptName) !== -1) {
   logger.warning("Not attaching missing or substituted font: " + fontsInProject[i].fontPostScriptName, jobTemplateHelperFile);
} else {
    fontsInProjectFiltered.push(fontsInProject[i]);
}
```

src/submission/JobTemplateHelper.jsx
viknith: Essentially, warnings are good, but I'm concerned about the job failing due to substituted fonts, and we should have a checkbox for customers on whether customers want substituted font jobs to pass or fail.

viknith:
**On Windows and MacOS for AE 24.6.4 and AE 25.1, I was able to confirm that the error messages popped up correctly:**

Screenshot 2025-02-02 at 3 30 50 PM
Screenshot 2025-02-02 at 3 32 26 PM

**And that the job attachments did not include substituted files with the missing font names**

Screenshot 2025-02-02 at 2 45 00 PM
Screenshot 2025-02-02 at 3 22 05 PM

And here was the job submission results when they were submitted, all failing.
Screenshot 2025-02-02 at 3 34 53 PM
Screenshot 2025-02-02 at 3 49 22 PM

Seems to be all in working order! Just need to address some of the comments from above, I think the one from Hao about letting customers know that their job will likely fail due to the missing font is important.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
eherozhao: I am trying to understand this behavior from customer perspective:
1. If I am a customer wanting to submit a job with missing custom font and I don't really care about the missing fonts, the render should succeeded.
2. If I am a customer wanting to submit a job with missing custom font and I want to make the change to update them before the job is rendered in the Deadline Cloud side.
In this way, I think we can have a warning pop up for customers that "are you sure you are going to submit a job with missing fonts?" with a yes and no ask. If yes, our job attachment should auto detect the replaced fonts with the replaced font name (not the one shown in the application)" If no, we just cancel the job submission.

I don't think we always fail customers missing fonts in the worker with no ask is a good customer experience, especially the worker has to run all steps before this failure, which seems wasting their money.

eherozhao (changes requested):
Do you mind also fixing the SonarQube and signing? Thanks!

cilevitz:
SonarQube issues have been resolved.

src/submission/JobTemplateHelper.jsx
cilevitz: Resolved in c9c25ca

src/submission/JobTemplateHelper.jsx
cilevitz: Missing or substituted fonts are now added to job attachments with the name of their substitution, e.g. `CyTextBold` substituted to `ArialMT` gives an attachment named `ArialMT`.

Missing fonts will not cause a job to fail on the farm, but will produce incorrect render output.

The user is warned when clicking Submit that they should install missing or substituted fonts to ensure correct render output.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
cilevitz: Font scan errors from Python are now surfaced in the AE UI.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
cilevitz: We added a confirmation dialog when submitting a composition containing missing or substituted fonts.

README.md
```diff
@@ -76,11 +76,11 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
...
 
-The submitter detects fonts used in the submitted composition and automatically adds them as job attachments on submission. These get installed on the worker before the render starts and get removed again when the job ends.
+Fonts used in the submitted composition are detected by the submitter and are automatically added as job attachments on submission. These get installed on the worker before the render starts and get removed when the job ends.
 Supported font types include: OpenType (`.otf`), TrueType (`.ttf`), and [Adobe Fonts](https://fonts.adobe.com/).
 Windows bitmap fonts (`.fon`) are only supported on Windows machines.
 
-If fonts are missing at render time, first check that they're installed (on the system or your user), and then check they're being included in the job attachments tab in the submitter.
+If fonts are missing or substituted at render time, then render tasks will fail. You must add any required fonts to your system or your user profile before opening a project that needs them. First check that the missing fonts are installed (on the system or your user), and then check they're being included in the job attachments tab in the submitter.
```
eherozhao: I thought we wouldn't fail the render job if the font was substituted if I understood the implementation correctly. Please confirm it.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
```diff
@@ -177,13 +184,24 @@ def uninstall_font(src_path, scope=INSTALL_SCOPE_USER):
 
         # Remove the fontname/filename from the registry
         fontname = get_font_name(dst_path)
-
-        with winreg.OpenKey(registry_scope, FONTS_REG_PATH, 0, access= winreg.KEY_SET_VALUE) as key:
-            winreg.DeleteValue(key, fontname)
+        
+        try:
+            with winreg.OpenKey(registry_scope, FONTS_REG_PATH, 0, access= winreg.KEY_SET_VALUE) as key:
```
viknith: Nit: `access= winreg.KEY_SET_VALUE` has extra space after equal sign

src/submission/JobTemplateHelper.jsx
```diff
@@ -348,9 +407,43 @@ function createFontFilename(fontLocation, fontPostScriptName) {
...
+    }
+
     return fontName;
 }
 
+/**
+ * Collects metadata from a given font and determines if it needs an override name.
+ * @return Object containing the override name or an error
+ **/
+function getFontNameOverride(fontObject, fontPostScriptName, fontLocation, oldLocation) {
```
viknith: nit: oldLocation is an unused variable

src/submission/JobTemplateHelper.jsx
```diff
@@ -310,11 +344,36 @@ function getLocationForFont(fontPostScriptName) {
...
 
+/**
+ * Gets the PostScript name of an installed font whose file name is fontFileName.
+ * @return The PostScript name of the font or null if the font was not found
+ **/
+function getPostScriptNameForFont(fontFileName) {
+    var fontPostScriptName = null;
+    try {
+        // Get user-installed fonts
+        var fontPaths = getFontPaths();
```
viknith: Not a blocker for this PR, but we should cache the output of getFontPaths because the output is the same each time

src/submission/SubmitBundle.jsx
```diff
@@ -92,6 +92,17 @@ function SubmitSelection(selection, framesPerTask) {
     var dependencies = findJobAttachments(rqi.comp); // list of filenames
     var compName = dcUtil.removeIllegalCharacters(rqi.comp.name);
 
+    if (app.fonts.missingOrSubstitutedFonts != "") {
+        var submitConfirmation = confirm(
```
viknith: Just tried submitting a job that had no missing or substituted fonts, all using just Amazon Ember and the GUI submitter CLI command was not reached and the submitter fails silently.

The root caused is happening around these lines of code.

Most likely it's because this variable is never created and set to true if there are no missing fonts. As a result, it defaults to false, returns, and then there's not notification to the customer. This needs to be set to default true initially and then have the if-check with confirmation. Requires code fix.

src/submission/JobTemplateHelper.jsx
```diff
@@ -140,46 +140,80 @@ function findJobAttachments(rootComp) {
...
+                    app.fonts.missingOrSubstitutedFonts.toString() + "\n\n" +
+                    "The font has been substituted with a different font by After Effects and will render as the substituted font instead.\n" +
+                    "Please install the font and reopen this project to ensure correct render output.";
+            }
+                
+            adcAlert(error_msg, false);
+        }
+        // Build a new array containing only attachable fonts.
+        var fontsInProjectFiltered = [];
+        for (i = 0; i < fontsInProject.length; i++) {
```
eherozhao: nit: do we need it to be `var i = 0` here?

src/submission/JobTemplateHelper.jsx
eherozhao: If I am using a unsupported ttc font type, what will it behave? I didn't see the notification that the font was substituted to the default font so I think it might be a missing part when the font was not supported (but not missing). Wanna hear about your thoughts on this.

src/submission/SubmitBundle.jsx
viknith: Tried to nest the second if statement under the first with missing fonts but that didn't work, AE ends up loading and then crashes from not responding. This is AE 25.1 on MacOS btw

src/submission/JobTemplateHelper.jsx
cilevitz: Submitting a project from Mac with a TTC font unsupported by Windows (and no other missing or substituted fonts at submit time) will result in no warnings from the submitter. The font will not be installed by font_manager.py on a Windows render node, as unsupported fonts cannot be installed. aerender will warn about missing fonts at render time. Render output will have substituted fonts.

README.md
cilevitz: Documentation has been updated in ea25a1f to indicate a choice to submit a job with missing or substituted fonts, even though the render output will not be correct.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1875,344 +1971,340 @@
...
+  }
+
+  if (typeof Date.prototype.toJSON !== "function") {
+    Date.prototype.toJSON = function () {
+      return isFinite(this.valueOf())
+        ? this.getUTCFullYear() +
+            "-" +
+            f(this.getUTCMonth() + 1) +
+            "-" +
+            f(this.getUTCDate()) +
```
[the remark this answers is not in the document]
cilevitz: This is a spurious error.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py
cilevitz: Resolved in ea25a1f

src/submission/JobTemplateHelper.jsx
cilevitz: Resolved in ea25a1f

src/submission/JobTemplateHelper.jsx
cilevitz: fontPaths are now cached in ea25a1f

src/submission/SubmitBundle.jsx
cilevitz: Resolved in ea25a1f

src/submission/JobTemplateHelper.jsx
cilevitz: `i` is reused from earlier in the function and doesn't need redeclaration. Changed `i` to `f` for clarity.

src/submission/JobTemplateHelper.jsx
```diff
@@ -310,11 +343,35 @@ function getLocationForFont(fontPostScriptName) {
...
+    try {
+        for (var path in fontPaths) {
+            if (path.indexOf(fontFileName) !== -1) {
+                // Found name that matches the given font's path
+                fontPostScriptName = fontPaths[path]["postscript_name"];
+                break;
+            }
+        }
+    } catch (e) {
+        logger.error(e.message, jobTemplateHelperFile);
```
eherozhao, line 365: I saw this error logged in my log file.
> 2025-02-10 14:42:22 - [ERROR   ]  JobTemplateHelper.json: JSON.parse
2025-02-10 14:42:22 - [ERROR   ]  JobTemplateHelper.json: Object of type Object found where a Number, Array, or Property is needed

I think it was from this line. Can you please check if this was parsed correctly?

eherozhao (changes requested):
image
I tried to submit a job for a scene with unsupported font. For the existing mainline version, the font was substituted by AE to some default font. However, using this PR branch, I didn't see the confirmation introduced from this PR asking about if I should submit the job if there is a missing or substituted fonts, also the job attachment was not having the substituted fonts. This will make the customer experience worse.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -91,6 +95,11 @@ def main():
             if "WARNING:After Effects warning" in line:
                 print(f"After Effects Warning: {line.strip()}", file=sys.stderr)
 
+                # Some warnings should be treated as errors, i.e. a fatal warning
+                if any([w in line for w in FATAL_WARNINGS]):
+                    print(f"After Effects Fatal Warning: {line.strip()}", file=sys.stderr)
+                    sys.exit(1)
```
eherozhao, line 101: I submitted a job with missing fonts, but this didn't stop my job rendered. Here is the log.
```
2025/02/11 13:23:17-08:00 After Effects Warning: WARNING:After Effects warning: Grouped_Message_{This project contains references to missing effects. Please install the following effects to restore these references.|||This project contains a reference to a missing effect. Please install the following effect to restore this reference.###@0 missing effects.}"Saber"
2025/02/11 13:23:18-08:00 WARNING:After Effects warning: Project has missing fonts.
2025/02/11 13:23:18-08:00 After Effects Warning: WARNING:After Effects warning: Project has missing fonts.
2025/02/11 13:23:18-08:00  
2025/02/11 13:23:18-08:00 rqindex 1num 1
2025/02/11 13:23:18-08:00 PROGRESS:  2/11/2025 9:23:18 PM: Starting composition "test_comp".
2025/02/11 13:23:18-08:00  
2025/02/11 13:23:18-08:00 PROGRESS:  Render Settings: Best Settings
```

viknith:
Since there's been no updates or responses to comments and it's stale, we'll go ahead and close this PR. If it receives updates, go ahead and reopen the PR.

PR #120, opened 2025-02-03
ci: add release publish workflow for ae

moorec-aws:
AE needs a release publish workflow to perform release but no longer needs to build an push to PyPI

AE is getting its own publish workflow that does not build or publish the python package.

AE does version release but no longer publishes to PyPI

Tested in a private development account

No

No

---

### DIFF
--- a/.github/workflows/relase_publish.yml
+++ b/.github/workflows/relase_publish.yml
@@ -1,0 +1,88 @@
+name: "Release: Publish"
+run-name: "Release: ${{ github.event.head_commit.message }}"
+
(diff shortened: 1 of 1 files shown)


eherozhao (commented):
Will this workflow create the zip file in the release tab?

moorec-aws:
> Will this workflow create the zip file in the release tab?

Yes the source is zipped and added to the release by default

eherozhao (approved):

epmog (approved):

PR #128, opened 2025-02-04
fix: fixed getPythonExecutable() function for submitter on MacOS 

viknith:
The getPythonExecutable() was using "where" to identify where python was located on MacOS when it should've used "which". This is because "where" is supported on zsh but not on bash, which is where the system call is made on by default. Additionally, when searching for `python3`, it would look for the string `python`, which is not as precise as it should be.

Replace "where" with "which" for calls made on MacOS. 

In addition to this fix, I refactored the code so that it minimized redundancies, has smoother error handling, and is cleaner to read. That's why the function has a lot of edits to it.

Cleaner code, and the get_user_fonts.py file can now be called on submissions from MacOS

Replace this line of code
```
var fontLocation = font.location || getLocationForFont(fontPostScriptName);
```
with this line
```
var fontLocation = getLocationForFont(fontPostScriptName);
```
This forces the python script and python executable finding handler to be called. I confirmed that the python executable was correctly found and used on MacOS AE 24.6.4 + 25.1 and Windows AE 25.

No

No

---


src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@ function getFontsFromFile() {
     return fontLocations;
 }
 
-/**
- * Checks that the system has Python installed and version >= 3
- * @return String with executable name corresponding to Python 3, or an empty string if not found
- **/
```
eherozhao, line 190: Is there any reason we delete the docstring?

src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@ function getFontsFromFile() {
...
-    try {
-        outputWhere = system.callSystem(findCommand);
-        if (outputWhere && outputWhere.indexOf(findSuccess) !== -1) {
-            pythonFound = true;
-            pythonExecutable = "python";
-        } else {
-            logger.warning("Couldn't find Python with executable name 'python'");
+    for (var i = 0; i < pythonExecutables.length; i++) {
+        // Search for python executable
+        var python = pythonExecutables[i];
```
eherozhao: nit: can we rename it to `pythonExecutable`?

src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@ function getFontsFromFile() {
...
+        var findCommand = "which " + python;
+        var findSuccess = "/" + python;
+        var os = $.os.toLowerCase();
+        if (os.indexOf("win") !== -1) {
```
eherozhao: we have both `if (os.indexOf("win") !== -1) {` and `if (os.indexOf("windows") !== -1) {` used. I would prefer us to keep it consistent.

src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@ function getFontsFromFile() {
...
-        outputWhere = system.callSystem(findCommand);
-        if (outputWhere && outputWhere.indexOf(findSuccess) !== -1) {
-            pythonFound = true;
-            pythonExecutable = "python";
-        } else {
-            logger.warning("Couldn't find Python with executable name 'python'");
+    for (var i = 0; i < pythonExecutables.length; i++) {
+        // Search for python executable
+        var python = pythonExecutables[i];
+        var findCommand = "which " + python;
```
eherozhao: I recalled that you mentioned `which deadline` would work for both os. Is that right?

src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@
...
-        // Python wasn't found on the path, try find python3
         try {
-            outputWhere = system.callSystem(findCommandPy3);
-            if (outputWhere && outputWhere.indexOf(findSuccess) !== -1) {
-                python3Found = true;
-                pythonExecutable = "python3";
-            } else {
-                logger.warning("Couldn't find Python 3 with executable name 'python3'");
+            var outputWhere = system.callSystem(findCommand);
+            if (!outputWhere || (outputWhere && outputWhere.indexOf(findSuccess) === -1)) {
```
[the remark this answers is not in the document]
eherozhao: have you test this if condition? based on this code scanning, I am wondering if this if condition will not work as expected.

src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@ function getFontsFromFile() {
...
+                logger.warning("Couldn't find Python with executable name '" + python + "'");
+                continue;
             }
         } catch (e) {
             logger.error(e.message, jobTemplateHelperFile);
             logger.debug("Where command output: " + outputWhere, jobTemplateHelperFile);
         }
+
+        // Python executable was found, verify Python version
+        try {
```
eherozhao, line 218: if we find python3 executable, do we still need to call this?

src/submission/JobTemplateHelper.jsx
```diff
@@ -184,94 +184,66 @@ function getFontsFromFile() {
...
+            logger.error(e.message, jobTemplateHelperFile);
+            logger.debug("Command output: " + output, jobTemplateHelperFile);
+        }
     }
 
-    var errorMessage = "";
-    if (!(pythonFound || python3Found)) {
-        logger.error("No Python found on the path", jobTemplateHelperFile);
+    // If reaching here, this means python version was too low or executable was not found
+    if (pythonFound) {
```
eherozhao: I think you want to see python was not found? or python is found but the version is not higher that 3? If so, this name need to be modified to be more reasonable.

eherozhao:
Could you point me a font type on macOS that I can use to test this change?

src/submission/JobTemplateHelper.jsx
viknith: No, `which deadline` is for Mac bash and `where deadline` is for Windows.

src/submission/JobTemplateHelper.jsx
viknith: Yes, just to confirm version and to also verify that the python file we found is actually an executable and not some dummy file named `python3`. Or if python3 is pointing a version of Python2 for a strange reason. Low risk, but this covers that.

src/submission/JobTemplateHelper.jsx
viknith: Python not found is in the else criteria, python found but was less than version 3 is in the pythonFound case. I'll rename pythonFound be clearer like `oldPythonVersionFound`.

src/submission/JobTemplateHelper.jsx
viknith: Accident, will add them back

src/submission/JobTemplateHelper.jsx
eherozhao: Can we just ignore this case? If customer doens't have Python or Python2, we just return a general one that you need to make sure you have python3 in your path.

viknith:
> Could you point me a font type on macOS that I can use to test this change?

On MacOS, I have not yet found a font type that we can use to test this change yet naturally, I've only been able to test this by doing the steps I defined in the PR description above. IIRC, this Python script was originally made due to there being trouble finding fonts on the Windows side of things, not Mac.

eherozhao (dismissed):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1083,89 +1083,54 @@ function getFontsFromFile() {
...
-
-    if (!pythonFound) {
-        // Python wasn't found on the path, try find python3
+    var pythonExecutables = ["python3", "python"];
+
+    for (var i = 0; i < pythonExecutables.length; i++) {
+        // Search for python executable
+        var pythonExecutable = pythonExecutables[i];
+        var findCommand = "which " + pythonExecutable;
+        var findSuccess = "/" + pythonExecutable;
```
npmacl, line 1092: does this not work without the forward/backward slash?

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1083,89 +1083,54 @@ function getFontsFromFile() {
...
+        }
+        var outputWhere = null;
         try {
-            outputWhere = system.callSystem(findCommandPy3);
-            if (outputWhere && outputWhere.indexOf(findSuccess) !== -1) {
-                python3Found = true;
-                pythonExecutable = "python3";
-            } else {
-                logger.warning("Couldn't find Python 3 with executable name 'python3'");
+            outputWhere = system.callSystem(findCommand);
```
npmacl, line 1100: Why doesn't this require the "cmd -c"?

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1083,89 +1083,54 @@ function getFontsFromFile() {
...
+            if (!outputWhere || outputWhere.indexOf(findSuccess) === -1) {
+                logger.warning("Couldn't find Python with executable name '" + pythonExecutable + "'");
+                continue;
             }
```
npmacl, line 1104: Don't we need to break if we did find the executable? Suppose `python3` was on path but not `python`. It seems like right now, we would see that `python3` is on path, then check if `python` is on path, and output a warning to logs, and exit the loop with `pythonExecutable` still be set to `python`.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1083,89 +1083,54 @@ function getFontsFromFile() {
...
-                    "\n" +
-                    "Please ensure that Python 3 or higher is installed correctly.";
-                adcAlert(errorMessage, true);
+        // Python executable was found, verify Python version
+        var output = null;
+        try {
+            output = system.callSystem(pythonExecutable + " --version");
+            if (output && output.indexOf("Python ") !== -1) {
+                var pythonVersion = parseInt(output.substring(output.indexOf(" ") + 1));
+                if (pythonVersion >= 3) {
```
npmacl, line 1116: do we want to enforce 3.9+, or is 3 enough here?

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1083,89 +1083,54 @@ function getFontsFromFile() {
...
-        logger.error(e.message, jobTemplateHelperFile);
-        logger.debug("Command output: " + output, jobTemplateHelperFile);
     }
 
-    return pythonExecutable;
+    // If reaching here, this means python version was too low or executable was not found
+    var errorMessage =
+        "Error: Couldn't find Python on the path.\n" +
+        "\n" +
+        "Please ensure that Python 3 or higher is installed correctly and added to your PATH.";
```
npmacl: This error message will be confusing if Python _**is**_ on the path but is not the right version. We should adjust the error message based on the cause.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1177,21 +1142,15 @@ function getFontPaths() {
...
-        var scriptFile = new File(scriptPath);
-        if (!scriptFile.exists) {
-            errorMessage =
-                "Error: Missing font script at " + scriptFile.fsName + "\n" +
-                "\n" +
-                "Please ensure that the Deadline Cloud Submitter is installed correctly.";
-        }
-    }
-    if (errorMessage) {
+            "Please ensure that the Deadline Cloud Submitter is installed correctly.";
```
npmacl, line 1153: nit: "ensure the submitter is installed correctly" is not a very helpful for root causing an error.

src/submission/JobTemplateHelper.jsx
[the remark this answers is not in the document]
viknith: Fixed!

dist/DeadlineCloudSubmitter.jsx
viknith: We figured out that the prefixing of the system calls is optional. callSystem() will make calls to Command Prompt on Windows or Bash on Mac as expected. To modify what shell environment is being called or to be specific, then you can do `cmd -c` or `powershell` etc.

dist/DeadlineCloudSubmitter.jsx
viknith: 3 is enough here, we're just trying to run a Python script to fetch font files. 3.9+ is required for the GUI submitter and if we want a specific validation behavior for that, then we should put that elsewhere, outside the scope of this PR.

dist/DeadlineCloudSubmitter.jsx
viknith: I guess it can work in both cases, will test and revise if so.

dist/DeadlineCloudSubmitter.jsx
viknith: I'll revise the first line of the error message to be
```
Error: Couldn't find Python 3 or higher on your PATH.\n
```

dist/DeadlineCloudSubmitter.jsx
viknith: If you take a look at the resulting file, the error message looks like this
```
errorMessage =
            "Error: Missing font script at " + scriptFile.fsName + "\n" +
            "\n" +
            "Please ensure that the Deadline Cloud Submitter is installed correctly.";
```
So that will convey what the problem is.

dist/DeadlineCloudSubmitter.jsx
viknith: No, this wouldn't work because if Python is not found, then we'll get an error says "python not found".
For example
```
viknith@7cf34df03377 ~ % which blah      
blah not found
```
So we need the slash to confirm that it was found.

npmacl (approved):

eherozhao (approved):

PR #131, opened 2025-02-06
feat: Grey out "Frames Per Task" if output is not an image sequence

viknith:
Customer are able to submit video or audio job submissions and also edit the Frames Per Task textbox, implying that it has some functionality for non-image-sequence jobs. This is not true, and editing that box would not change anything.

Disable the box if a non-image-sequence job is selected for submission

Better UI for customers to minimize any confusion

Verified on MacOS and Windows AE 25.1 and AE 24.6.4 that when a video output composition was selected, the textbox was disabled, and when I switched to a composition with image sequence output, the textbox re-enabled. I also changed the outputs directly, and verified that upon refresh, the textbox would enable or disable depending on the selected output. Submitted jobs on MacOS and Windows on AE25.1 and they succeeded.

The selected composition here has mp3 output. The text box is disabled
Screenshot 2025-02-06 at 1 59 39 PM

Successful jobs
Screenshot 2025-02-06 at 5 43 39 PM

https://github.com/user-attachments/assets/073d7a44-1297-47fc-8796-d715f9e9ab9d

No

### Was this documented

No

---


src/ui/SubmitterUI.jsx
```diff
@@ -53,26 +53,51 @@ function buildUI(thisObj) {
...
+        // after submit button is clicked, so this is a sufficient for now
+        if (rqi.numOutputModules == 1) {
+            var outputModule = rqi.outputModule(1).file;
+            if (outputModule != null) {
+                const outputFile = outputModule.name;
+                const regex = new RegExp('\\b' + "%5B#####%5D" + '\\b', 'g');
+                const outputFileNameNoRegex = outputFile.replace(regex, "[#####]");
+                const lastIndex = outputFileNameNoRegex.lastIndexOf(".");
+                const extension = outputFileNameNoRegex.substring(lastIndex + 1);
+                return isImageOutput(extension);
```
eherozhao, line 88: We have this code snippet existing https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/src/submission/SubmitBundle.jsx#L183-L192

Can we make it something reusable?

viknith: At most I can make a helper function that gets the extension of a file name, but refactoring the regex will be more difficult until there's a more consistent usage of how file names are extracted. Let me see what I can do.

eherozhao (dismissed):

src/ui/SubmitterUI.jsx
```diff
@@ -120,23 +142,20 @@ function buildUI(thisObj) {
...
+    updateList();
 
     refreshButton.onClick = function() {
         updateList();
     }
 
-    submitterPanel.addEventListener('click', function() {
-        updateList();
-    }, true);
-
```
viknith, line 139: Removed this because it's redundant code, since that's what the refresh button should do

npmacl:
Why not not just remove the Frames per task box if submitting a video? At least to my eyes I would just interpret a grayed out number as a suggested default value or fixed at 10.

viknith:
Good question! We had given the submitter to people to test and two users requested that it should be greyed out. Let me speak with Hao to get further confirmation on whether it should be removed instead of greyed out.

viknith:
Yes, we want to grey out the box, not remove it. Some of the points made were that eventually we should be able chunking on videos too and in the short-term it conveys that we can't chunk videos yet, and it's better to have consistent features regardless of what is being submitted.

viknith:
After further discussion, we agreed to add text replacement in the greyed-out textbox so that customers can see why it's greyed out. Will push a revision!

npmacl (approved):

eherozhao (approved):

PR #134, opened 2025-02-10
chore: add py as a python executable option to cover Windows Python installation

viknith:
When Python is installed on Windows, the name of the executable can also be `py`, and its possible they won't have python or python3, but just py.

Add Py to the list of executables to check for.

Better handling for edge case customers

Not tested since it's a minor one-liner change.

No

No

---


npmacl (approved):

eherozhao:
It can be tested in windows by reinstalling python without adding it to env var. Not a blocker for the PR.

eherozhao (approved):

PR #139, opened 2025-02-15
fix: in macOS make GUI submitter independent on AE submitter to allow end users to keep using AE during assets upload

### LINKED ISSUE #136: Feature request: Submit without freezing After Effects GUI
### Describe the problem

While the submission is hashing and uploading all job attachments, the After Effects GUI is frozen. It'd be nice if we just performed the upload in the background instead so customers can do other tasks.

After Effects tends to have large job attachments because its dependencies are often large video files.

### Proposed Solution

Open the submitter in a separate process that doesn't block the GUI or add a feature to the Deadline Monitor (or other component) to manage uploads and downloads for us.

### Example Use Cases

When file sizes are large, it's cumbersome to keep after effects open for the entire duration of the upload. In our specific case, we had 200 GB of files that would take many hours to upload and the artist wanted to be able to work on other things while the upload was going.

eherozhao:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/136

GUI submitter is called in AE submitter which blocked AE submitter to be used if GUI submitter is still running or uploading assets.
In this PR, we found a way to open a Terminal in macOS to run the command line to free the AE application for end users to keep using. We will still need to figure out how to resolve this in Windows env.

This can help end users to keep using the application during the job assets upload.

I tested it in both Windows and macOS env and successfully submit the job.

N/A
---


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1624,8 +1624,10 @@ function SubmitSelection(selection, framesPerTask) {
         // Execute the command using a bash in the interactive mode so it loads the bash profile to set
         // the PATH correctly.
         var shellPath = $.getenv("SHELL") || "/bin/bash";
-        submitScriptContents = shellPath + " -i -c '" + cmd + "'";
-        output = system.callSystem(submitScriptContents + ' || echo "\nERROR CODE: $?"');
+        var cmd =
+            'deadline bundle gui-submit \\\"' + bundle.fsName + '\\\" --output json --install-gui --submitter-name \\\"After Effects\\\"';
+        submitScriptContents = shellPath + " -i -c \\\"" + cmd + "\\\" && exit";
+        output = system.callSystem('osascript -e \'tell application "Terminal"\' -e \'do script "' + submitScriptContents + '\"\'' + ' -e \'end tell\' > /dev/null');
```
viknith, line 1630: is there any output error logging from this system call that we can capture in the output rather than outputting to /dev/null?

viknith (approved):

joel-wong-aws:
What happens if assets are modified while the upload is in progress? e.g. if a user saves their comp while it is being uploaded?

viknith:
> What happens if assets are modified while the upload is in progress? e.g. if a user saves their comp while it is being uploaded?

This issue could still happen without this PR because customers are only limited from interacting with After Effects application, not from interacting with their machine as a whole. It should be tested though

npmacl (approved):

PR #159, opened 2025-03-27
fix: switch Deadline CLI from Terminal window call to shell script call to avoid Terminal window popup on Mac

viknith:
When submitting AE jobs, a separate Terminal window opens to call the Deadline GUI-submitter CLI in a separate process so that artists can still interact with AE while looking at the submitter window.

Switch from calling the Terminal application to running shell script directly

Mac AE submitter experience is better

Submitted jobs, the AE submitter opens and the Terminal window is not showing up

Yes

N/A

N/A

No

No

---


MarinMaksutaj (dismissed):

npmacl (dismissed):

viknith:
Not merging yet, the Terminal window is still popping up, not sure why, investigating further.

npmacl (approved):

viknith:
Updated PR description with notes, switching from a Terminal application call to a shell script call gets rid the Window popping up

MarinMaksutaj (approved):

PR #162, opened 2025-04-04
fix: added Python validation check on job submission

### LINKED ISSUE #153: Bug: Submitter silently fails if python is not installed
### Describe the bug

When submitting from a machine that doesn't have a python installation, the submit button brings up a command prompt window that disappears a split second later. No error is present.

### Expected Behaviour

We should at the very least show an error. In this case, the submitter was installed with the installer from the Deadline Cloud downloads page, so it should have also handled installing python for us if possible.

### Current Behaviour

A command prompt window briefly flashes.

### Reproduction Steps

Spin up a new Windows machine and install After Effects on it. Run the Deadline Cloud submitter installer. Try to submit.

### Environment

At minimum:
1. Operating system: Windows
2. Version of After Effects: 2025
3. Version of this package: The version available in DeadlineCloudSubmitter-2025.02.06.0-windows-x64-installer.exe

viknith:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/153

Customers were reporting that they were not getting proper error messaging when Python was not installed or added to PATH

Added additional python validation check on job submission so that job submission cannot proceed and fails early until it is validated that Python is installed and added to the customer's PATH env variable. This prevents having to add duplicate python validation in other places and centralizes it in one area.

Better error handling and messaging for customers

Updated the scripts in AE, uninstalled Python, then attempted to submit a job, and verified the popup appeared. Then I reinstalled Python and added it to my PATH and attempted job submission, which passed!

N/A

N/A

N/A

No

No

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -2412,13 +2412,15 @@ function buildUI(thisObj) {
 
     var submitButton = controlsGroup.add("button", undefined, "Submit");
     submitButton.onClick = function() {
-        const multiFrameRendering = mfrCheckBox.value ? "ON" : "OFF";
-        var maxCpuUsagePercentage = undefined;
-        if (mfrCheckBox.value) {
-            maxCpuUsagePercentage = parseInt(maxCpuUsagePercentageTextBox.text)
+        if (getPythonExecutable()) {
+            const multiFrameRendering = mfrCheckBox.value ? "ON" : "OFF";
+            var maxCpuUsagePercentage = undefined;
+            if (mfrCheckBox.value) {
+                maxCpuUsagePercentage = parseInt(maxCpuUsagePercentageTextBox.text)
+            }
+            SubmitSelection(list.selection, parseInt(framesPerTaskTextBox.text), multiFrameRendering, maxCpuUsagePercentage);
+            list.selection = null;
         }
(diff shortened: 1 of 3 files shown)


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -2412,13 +2412,15 @@ function buildUI(thisObj) {
...
-            maxCpuUsagePercentage = parseInt(maxCpuUsagePercentageTextBox.text)
+        if (getPythonExecutable()) {
+            const multiFrameRendering = mfrCheckBox.value ? "ON" : "OFF";
+            var maxCpuUsagePercentage = undefined;
+            if (mfrCheckBox.value) {
+                maxCpuUsagePercentage = parseInt(maxCpuUsagePercentageTextBox.text)
+            }
+            SubmitSelection(list.selection, parseInt(framesPerTaskTextBox.text), multiFrameRendering, maxCpuUsagePercentage);
+            list.selection = null;
         }
```
eherozhao, line 2423: If we would like to do this check, I think we should throw the error in a else block to notify customers that their python executable is not set as expected and pointing them to any possible reference to unblock the issue.

viknith: The error message happens in the getPythonExecutable function call

viknith: The getPythonExectuable function has error handling built into that because it's used in other places.

npmacl (approved):

eherozhao (approved):

PR #187, opened 2025-05-02
feat: add support for After Effects submitter user install on Mac & Windows

viknith:
The Deadline Cloud submitter for After Effects has inherently always been a system install and has required customers to have elevated permissions to install the submitter. 

After Effects supports allowing the customer to select the script they would like to run. Since the AE submitter is basically a script (technically a script panel though) it can be executed by After Effects when the user selects it via file system exploration. As a result, we'll download the AE submitter under the DeadlineCloudSubmitter folder under the corresponding user folder which won't require elevated permissions! 

Customers will be unblocked from doing proper user installs on Windows and Mac! The customer will have a slightly different procedure to running the submitter, but the documentation has been updated for that. The downside of a proper user install for now is that the submitter is no longer dockable, but the tradeoff is acceptable for now in order to support user installs.

Ran AE dev installer on Windows and Mac and verified that the submitter was installed to the correct locations and were properly uninstalled as well! The user installation would install the submitter scripts to a non-privileged location under the DeadlineCloudForAfterEffectsSubmitter folder and the system installation would install the submitter scripts to the privileged location under the AE Plug-ins folder.

I then submitted render jobs from Mac and Windows for System and User installs and verified that all renders passed successfully!

N/A

Ran installer tests on Mac and Windows at the user level and system level, all passing!

Yes, README.md instructions have been updated with updated installation and usage instructions and DEVELOPMENT.md has been updated with new instructions on running tests.

No!

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,25 +66,25 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
 
 1. Next step is to install the After Effects submitter. We recommend choosing the submitter installer approach to receive both Deadline CLI and your selected integrated submitters. However, if you want to update your submitter with the latest code pushed to this repository, choose the manual installation approach, but ensure you have Deadline CLI already installed via pip or submitter installer beforehand (see instructions above under Prerequisites for info on this).
 
-*NOTE: if the After Effects submitter is not installed as a system install, the submitter will not be a dockable but will be a standalone submitter window.*
+*NOTE: If you install the After Effects submitter as a user install, the submitter will be a standalone submitter window rather than a dockable panel.*
 
    - **Submitter Installer Approach**
       - First, download the Deadline Cloud Submitter installer by following [Step 1: Install the Deadline Cloud Submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html#submitter-installation).
-      - Next, if you are running a System Install, right-click the installer and choose `Run as Admin`. Otherwise, run the installer regularly for a User Install.
-      - Follow the prompts and select which submitters you would like to install. Here are the following OS-specific submitters paths depending on your install approach:
-         - Windows default path for AE System Install: `C:\Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`.
-         - Mac default path for AE System Install: `/Applications/Adobe After Effects <version>/Scripts/ScriptUI Panels`
-         - Windows default path for AE User Install: `C:\Users\<user>\DeadlineCloudSubmitter\Submitters/AfterEffects\AE<version>`
-         - Mac default path for AE User Install: `/Users/<user>/DeadlineCloudSubmitter/Submitters/AfterEffects/AE<version>`
-      - If chosing a User Install, save the path where you stored the JSX submitter script.
+      - Next, if you are running a System Install on Windows, right-click the installer and choose `Run as Admin`. Otherwise, run the installer regularly for a User Install.
+      - Follow the prompts and select which submitters you would like to install. Here are the following OS-specific default submitters paths depending on your install approach:
+         - **Windows with System Installation**: `C:\Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`
+         - **Windows with User Installation**: `C:\Users\<user>\DeadlineCloudSubmitter\Submitters/AfterEffects\AE<version>`
+         - **macOS with System Installation**: `/Applications/Adobe After Effects <version>/Scripts/ScriptUI Panels`
+         - **macOS with User Installation**: `/Users/<user>/DeadlineCloudSubmitter\Submitters/AfterEffects\AE<version>`
+      - If choosing a User Install and you provide a custom install path, be sure to save that path for later reference.
 
    - **Manual Installation Approach**
       - Scroll to the top of this repository's Github page, and click on the green Code button. In the drop-down, select `Download ZIP`.
       - Unzip the .zip file and navigate to the `dist` folder in the downloaded repository. The submitter files are `DeadlineCloudSubmitter.jsx` and the `DeadlineCloudSubmitter_Assets` folder.
       - For System Install that requires Admin permission, you need to move the submitter files to the ScriptUI Panels folder of After Effects. This will require Admin permission.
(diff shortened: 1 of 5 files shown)


epmog (dismissed):
Change looks reasonable - approved!

My only non-blocking comment would be that the different versions for the user installation aren't needed since it's the exact same code, but that's pretty minor.

viknith:
Agreed, but I figured we should just have the mechanism baked in just in-case there is any version specific code files that we'd want to include/exclude.

README.md
```diff
@@ -65,21 +65,28 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
...
-         - Copy `DeadlineCloudSubmitter.jsx` and the `DeadlineCloudSubmitter_Assets` folder from the `dist` folder to ScriptUI Panels folder under `Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels` within your After Effects installation.
-   - MacOS:
-      - **Manual Installation Approach**:
-         - Scroll to the top of this repository's Github page, and click on the green Code button. In the drop-down, select `Download ZIP`.
-         - Unzip the .zip file and navigate to the `dist` folder in the downloaded repository.
-         - Copy `DeadlineCloudSubmitter.jsx` and the `DeadlineCloudSubmitter_Assets` folder from the `dist` folder to ScriptUI Panels folder under `Applications/Adobe After Effects <version>/Scripts/Script UI Panels` within your After Effects installation.
-      - **Submitter Installer Approach**: No support yet, will update with instructions once we add support for it!
+
+*NOTE: if the After Effects submitter is not installed as a system install, the submitter will not be a dockable but will be a standalone submitter window.*
```
eherozhao: nit:  change to positive expression "If you install the After Effects submitter as an user install"

README.md
```diff
@@ -65,21 +65,28 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
...
+
+   - **Submitter Installer Approach**
+      - First, download the Deadline Cloud Submitter installer by following [Step 1: Install the Deadline Cloud Submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html#submitter-installation).
+      - Next, if you are running a System Install, right-click the installer and choose `Run as Admin`. Otherwise, run the installer regularly for a User Install.
+      - Follow the prompts and select which submitters you would like to install. Here are the following OS-specific submitters paths depending on your install approach:
+         - Windows default path for AE System Install: `C:\Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`.
+         - Mac default path for AE System Install: `/Applications/Adobe After Effects <version>/Scripts/ScriptUI Panels`
```
eherozhao: I think we use macOS in other places, can we keep it consistent? Also, can we have it structured as 
- Windows with System Installation:
- Windows with User Installation: 
- macOS with System Installation:
- macOS with User Installation:

README.md
```diff
@@ -65,21 +65,28 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
...
+         - Windows default path for AE User Install: `C:\Users\<user>\DeadlineCloudSubmitter\Submitters/AfterEffects\AE<version>`
+         - Mac default path for AE User Install: `/Users/<user>/DeadlineCloudSubmitter/Submitters/AfterEffects/AE<version>`
+      - If chosing a User Install, save the path where you stored the JSX submitter script.
+
+   - **Manual Installation Approach**
+      - Scroll to the top of this repository's Github page, and click on the green Code button. In the drop-down, select `Download ZIP`.
+      - Unzip the .zip file and navigate to the `dist` folder in the downloaded repository. The submitter files are `DeadlineCloudSubmitter.jsx` and the `DeadlineCloudSubmitter_Assets` folder.
+      - For System Install that requires Admin permission, you need to move the submitter files to the ScriptUI Panels folder of After Effects. This will require Admin permission.
+         - Windows: Move them to `C:\Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`
+         - Mac: Move to `/Applications/Adobe After Effects <version>/Scripts/ScriptUI Panels`
```
eherozhao: Nit: same comment as before. prefer macOS.

README.md
```diff
@@ -90,10 +97,12 @@ The submitter includes a folder `DeadlineCloudSubmitter_Assets` and a file `Dead
...
-1. You can also specify multi-frame rendering with your job submission. If you do, you can also specify the max percentage of CPU usage you wish to allocate towards rendering in case you would like to limit it to allow other background applications or processes to run smoothly. For more information about multi-frame rendering, visit Adobe's website [here](https://helpx.adobe.com/after-effects/using/multi-frame-rendering.html) for more
+1. To open the Deadline Cloud Submitter Panel, there is a different approach depending on whether it's a User Install or a System Install.
+   - **System Install**: Open submitter by clicking **Window > DeadlineCloudSubmitter.jsx**.
+   - **User Install**: Open submitter by clicking **File > Scripts > Run Script File** and navigate to where the `DeadlineCloudSubmitter.jsx` file is located and select it to run the submitter. If the submitter is closed, reopen it easily by clicking **File > Scripts > Recent Script Files** and pick the `DeadlineCloudSubmitter.jsx` file that was previously run.
```
eherozhao, line 102: two questions:
1. Have you tried submitted the job after this installation?
2. Why not suggest customer to copy/paste the folder to their user installed path?

viknith: 1. As mentioned in PR description, all install approaches have been tested with the local dev installer with job submission and they pass. 
2. The copy-paste approach is mentioned above for customers that want to do a "Manual install". For using the Submitter Installer, there are separate instructions for the user installation so that it doesn't require copying files to a privileged path.

eherozhao: > there are separate instructions for the user installation so that it doesn't require copying files to a privileged path.

why do we introduce another way instead of asking customers to copy/paste the folders? Is there any reason blocking customers to do this with submitter installer user installation?

viknith: If we have a change that is currently in mainline but has not been released to the submitter installer, this provides an alternative way for customers to get the changes faster so that they don't have to wait for a formal release. This is applicable when we've put up feature changes to mainline but we haven't done a release yet.

eherozhao (approved):

MarinMaksutaj (approved):

PR #189, opened 2025-05-08
chore(installer): modify AE submitter install location to default to user install location on macOS

viknith:
The user/system choice selection is disabled on OSX for now, which means the installscope variable is not being set.

Default to user install approach on OSX without checking the installscope variable because it's not being set in the Mac installer currently.

Enables user install on Mac

Ran submitter install on Mac and Windows, both worked for user install and Windows system install.

N/A

All tests passed on Windows and Mac

Yes

No, this hasn't been released yet.

---


MarinMaksutaj (approved):

miabatta (approved):
Looks good! Thanks Vikram 💯

PR #192, opened 2025-05-13
fix: check folder permissions during installation

yuanmich2:
Fixes: an issue where submissions were failing because access to the temp folder was restricted.

Some users didn't have access to the directory that Folder.temp.fsName pointed to. This meant that their submitter couldn't write the required job bundle files.

I added a utility function that checks folder permissions and automatically falls back to alternate temp folder paths. I then made sure to replace instances of Folder.temp with the utility function.

Submissions are more robust.

I ran submissions on my own farm.

Yes

None of this really needs to be updated because the user experience doesn't change except that some people who were unable to render before can now render.

No

---

### DIFF
--- a/src/utils/Utils.jsx
+++ b/src/utils/Utils.jsx
@@ -324,9 +324,81 @@ function __generateUtil() {
         /**
          * Return File instance from temporary directory with the given name.
          */
-        var _tempFilePath = normalizePath(Folder.temp.fsName + "/" + fileName);
+        var _tempFilePath = normalizePath(getTempFolder() + "/" + fileName);
         var _tempFile = File(_tempFilePath);
         return _tempFile;
+    }
+
+    var _cachedTempFolder = "";
+
+    function getTempFolder() {
+        /**
+         * Fetches a temporary directory that has read/write access. Defaults to Folder.temp.
+         * Once a working temporary directory is found, the result is cached.
+         */
+        if (_cachedTempFolder != "") {
+            return _cachedTempFolder;
+        }
+
+        //Build list of alternate paths to try
+        var altPaths = [Folder.temp.fsName];
+        var testFileSuffix;
+        if (system.osName == "MacOS") {
+            altPaths.push("~/.deadline/DeadlineCloudAETemp");
+            altPaths.push(Folder.myDocuments.fsName + "/DeadlineCloudAETemp");
+            testFileSuffix = "/testFile.txt";
+        } else {
+            altPaths.push(Folder.userData.fsName + "\\DeadlineCloudAETemp");
+            altPaths.push(Folder.myDocuments.fsName + "\\DeadlineCloudAETemp");
+            testFileSuffix = "\\testFile.txt";
+        }
+
+        //Test every path in our list by creating a test file
+        for (var altPath in altPaths) {
+            var folder = new Folder(altPath);
+
+            //Create the path if it does not already exist
+            folder.create();
+            if (folder.exists) {
+                continue;
+            }
+
+            //List all files and check for errors
+            //Not having list permissions is a common source of errors
+            var existingFiles = folder.getFiles();
+            if (!folder.error) {
+                continue;
+            }
+
+            //Delete any existing files in the path
+            for (var existingFile in existingFiles) {
+                existingFile.remove();
+            }
+
+            //Create and write to a test file to make sure we have write permissions
+            var file = new File(folder.fsName + testFileSuffix);
+            file.open("w");
+            file.writeln("test");
+            file.close();
+            if (!file.exists || file.error) {
+                continue;
+            }
+            file.remove();
+
+            //Cache the folder that passed all the checks so we don't have to repeat the checks
+            _cachedTempFolder = folder.fsName;
+            break;
+        }
+
+        //If no alternate path was found, alert the user and return the default temp path
+        if (_cachedTempFolder == "") {
+            adcAlert("Error: No valid temporary directory found. Check permissions to one of these paths: " + altPaths.join(", "));
+
+            //we don't want to cache the result on a failure, just return the default value
+            return Folder.temp.fsName;
+        }
+
+        return _cachedTempFolder;
     }
 
     function wrappedCallSystem(cmd) {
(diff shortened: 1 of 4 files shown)


src/utils/Utils.jsx
```diff
@@ -324,11 +324,64 @@ function __generateUtil() {
...
+            file.writeln("test");
+            file.close();
+            if (file.exists && file.error == "") {
+                file.remove();
+                _cachedTempFolder = folder.fsName;
+                break;
+            }
+        }
+
+        //Will be an empty string if no directory with write permissions was found
```
eherozhao: nit: I think having some logs here can help.

src/utils/Utils.jsx
```diff
@@ -324,11 +324,64 @@ function __generateUtil() {
         /**
          * Return File instance from temporary directory with the given name.
          */
-        var _tempFilePath = normalizePath(Folder.temp.fsName + "/" + fileName);
+        var _tempFilePath = normalizePath(getTempFolder + "/" + fileName);
```
viknith: Potential nit: Should this be getTempFolder()? Or does just specifying the plain function name work here?

viknith (dismissed):

eherozhao (dismissed):

src/utils/Utils.jsx
```diff
@@ -324,11 +324,64 @@ function __generateUtil() {
...
+                file = new File(folder.fsName + "\\testFile" + Math.random().toString());
+            }
+            file.open("w");
+            file.writeln("test");
+            file.close();
+            if (file.exists && file.error == "") {
+                file.remove();
+                _cachedTempFolder = folder.fsName;
+                break;
+            }
```
viknith: Can you expand on the purpose of this code block? Some comments would be nice.

src/utils/Utils.jsx
yuanmich2: Not a nit. That's def a typo

src/utils/Utils.jsx
```diff
@@ -324,11 +324,83 @@ function __generateUtil() {
...
+        /**
+         * Fetches a temporary directory that has read/write access. Defaults to Folder.temp.
+         * Once a working temporary directory is found, the result is cached.
+         */
+        if (_cachedTempFolder != "") {
+            return _cachedTempFolder;
+        }
+
+        //Build list of alternate paths to try
+        let altPaths = [Folder.temp.fsName];
```
viknith: `let` keyword causes issues in ExtendScript, switch to `const` or `var` for all instances of `let`

src/utils/Utils.jsx
```diff
@@ -324,11 +324,83 @@ function __generateUtil() {
...
+        for (let altPath in altPaths) {
+            let folder = new Folder(altPath);
+
+            //Create the path if it does not already exist
+            if (!folder.exists) {
+                folder.create();
+            }
+            if (!folder.exists) {
+                continue;
+            }
```
viknith: Should be shrunk to just creating the folder (since creating an already existing folder shouldn't throw an error, but double check this assumption) and then check if it exists. One less if statement

src/utils/Utils.jsx
```diff
@@ -324,11 +324,83 @@ function __generateUtil() {
...
+
+            //List all files and check for errors
+            //Not having list permissions is a common source of errors
+            let existingFiles = folder.getFiles();
+            if (folder.error != "") {
```
viknith: Nit: !folder.error, same for all other instances of this.

src/utils/Utils.jsx
```diff
@@ -324,11 +324,83 @@ function __generateUtil() {
...
+                continue;
+            }
+
+            //Delete any existing files in the path
+            for (let existingFile of existingFiles) {
+                existingFile.remove();
+            }
+
+            //Create and write to a test file to make sure we have write permissions
+            let file = new File(folder.fsName + testFileSuffix);
```
viknith: Nit: consider using writeFile helper if possible: https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/8dd22edc46fb9dcc57c60b65c378b5a3e36487e2/src/utils/Utils.jsx#L12-L19

viknith: Also, is there a forward/backward slash missing here?

src/utils/Utils.jsx
```diff
@@ -324,11 +324,83 @@ function __generateUtil() {
...
+            }
+            file.remove();
+
+            //Cache the folder that passed all the checks so we don't have to repeat the checks
+            _cachedTempFolder = folder.fsName;
+            break;
+        }
+
+        //If no alternate path was found, prompt the user for a valid path
+        if (_cachedTempFolder == "") {
```
viknith, line 394: Just to confirm, what's the user experience like when you reach this else case? Does the GUI submitter still pop up or does the execution of the script end?

src/utils/Utils.jsx
yuanmich2: Thanks for the heads up! I made the change to let last minute because the github bot told me to. Probably should have tested before I pushed.

src/utils/Utils.jsx
yuanmich2: The else case is the normal case. This if statement is just to catch what happens when no usable folder was found. I just updated it to print a better message and I added comments.

src/utils/Utils.jsx
yuanmich2: I need a reference to the file object to get errors, so I'm not sure that the writeFile helper would work here.

src/utils/Utils.jsx
yuanmich2: fixed

src/utils/Utils.jsx
yuanmich2: The testFileSuffix starts with a slash, which slash is determined by the OS.

src/utils/Utils.jsx
yuanmich2: Just tested and you were right, it doesn't throw an error, so I'm getting rid of the first if statement.

viknith (approved):

eherozhao (approved):

PR #204, opened 2025-06-19
fix: include setting minor version for CondaPackages field and added version mismatch warnings

### LINKED ISSUE #203: Bug: Conda does not pick subversions, which results in compatability issues if submitting from an earlier version of AE
### Describe the bug

If you submit a version using AE 2025.1, the submitter just lists 25, and the result is the Conda Launch defaults to 2025.2.2.

This results in the render hanging for a long period of time- it never finishes unless you cancel the job.

If you specify the subversion in the submitter- e.g. 25.1 - then it works.

### Expected Behaviour

The submitter should use subversions to ensure compatability - e.g. a 2025.1 job specifies 25.1 in the submitter conda package

### Current Behaviour

It defaults to 25 and old versions of AE fail to render.

### Reproduction Steps

Submit a job from 2025.1 to the current farm - it will not work.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; macOS Sequoia 15.1.1; etc.)
2. Version of After Effects
3. Version of this package

viknith:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/203

If a customer submits a job from AE 25.1 locally to their service-managed fleet, the AE version is defaulting to 25, which means our fleet will get the latest conda package of AE we have, which is AE 25.2, which results in a version mismatch and causes renders to fail.

Add minor version to the conda package parameter field to prevent version mismatch. If the customer has a version of AE we don't support, we just default to the major version setting and issue a warning.

Removes the annoying experience of setting the minor version every time an artist wants to submit a job.

On MacOS, I attempted submitting jobs from AE 24.6, AE 25.1, AE 25.2 and AE 25.3. I verified that the minor version was set correctly for all of these versions except 25.3, where I got a warning message and the version just defaulted to 25.

Supported version submissions:
Screenshot 2025-06-19 at 11 06 03 AM
Screenshot 2025-06-19 at 11 08 53 AM
Screenshot 2025-06-19 at 4 02 19 PM

And here's the AE 25.3 submission, which we don't support:
Screenshot 2025-06-19 at 3 42 45 PM
Screenshot 2025-06-19 at 3 42 56 PM

Then I did the same on Windows for AE 25.2 and AE 25.3 (skipped 24.6 and 25.1 because it would've been redundant) and it worked the same way

Yes

Yes, in DEVELOPMENT.md

Nope

---

### DIFF
--- a/src/utils/Utils.jsx
+++ b/src/utils/Utils.jsx
@@ -760,7 +760,7 @@ function __generateUtil() {
 
     function getCompatibleAEVersion() {
         /* Return compatible After Effects version for job submission.
-         * Warns if current version is not officially supported.
+         * Warns if current version is not officially supported on service-managed fleets.
          * Returns the version as float.
          */
         const currentVersion = getAEVersion();
@@ -772,7 +772,7 @@ function __generateUtil() {
         // Show warning if version is not supported
         adcAlert(
             "Warning: Your After Effects version " + currentVersion +
-            " is not officially supported. Supported versions are: 24.6, 25.1, and 25.2. " +
+            " is not officially supported on service-managed fleets. Supported versions are: " + SUPPORTED_VERSIONS.join(", ") + ". " +
             "This may result in compatibility issues or failed jobs.",
             false
         );
--- a/src/utils/Utils.jsx
+++ b/src/utils/Utils.jsx
@@ -1,4 +1,5 @@
 var scriptFolder = Folder.current.fsName;
+const SUPPORTED_VERSIONS = [24.6, 25.1, 25.2];
 
 function readFile(filePath) {
     var f = new File(filePath);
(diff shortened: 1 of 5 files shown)


src/utils/Utils.jsx
```diff
@@ -752,11 +753,33 @@ function __generateUtil() {
...
+        const currentVersion = getAEVersion();
+
+        if (SUPPORTED_VERSIONS.indexOf(currentVersion) !== -1) {
+            return currentVersion;
+        }
+
+        // Show warning if version is not supported
+        adcAlert(
+            "Warning: Your After Effects version " + currentVersion +
+            " is not officially supported. Supported versions are: 24.6, 25.1, and 25.2. " +
```
eherozhao: nit: can we generate this from SUPPORTED_VERSIONS to make it more extendable?

eherozhao: "is not officially supported" -> we need to highlight this is not supported for Service-Managed Fleet Conda Package. We can also link our queue environment supported conda package there.

viknith: Missed that, will do!

viknith: Good call, will fix.

eherozhao (approved):

jaikish (approved):

PR #207, opened 2025-06-23
feat!: Support multiple composition submissions in one job

### LINKED ISSUE #206: Feature request: Submit multiple compositions at once
### Describe the problem

It is a common workflow to have multiple AE comps that you want to render at the same time.
We should be able to submit them in one go.

### Proposed Solution

Support each composition as a separate Step in DeadlineCloud.
Each comp needs to be able to have their own settings (chunk size, MFR, etc).

### Example Use Cases

It is a common workflow to have multiple AE comps that you want to render at the same time.

Ahuge:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/206

If you wished to submit multiple compositions at once as separate steps, this was not currently possible.
Refactor and add support to the submitter to allow for this feature to be possible.

I've attempted to keep the commits fairly granular so that reviewers can follow the train of thought throughout the process of the changes. I can squash changes as needed.
The logic around the job template has gotten more complex as we now have to generate a job template.
We have tried to make this easier by providing "fragments" of the templates that are used to generate the final template.
- [x] Testing using SMF from a Windows AE 25.3 client.
- [x] Testing using SMF from a Windows AE 25.2 client.
- [x] Tested using SMF from a MacOS AE 25.2 client
- [x] Testing using SMF from a Windows AE 24.6 client.
Yes

Not documented.

 Yes. The code structure has changed enough that this is breaking. We are also now generating the job bundles so this is breaking

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ node_modules
 .idea
 *.code-workspace
 .vscode
+jsconfig.json
 
 # macOS
 .DS_Store
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1810,7 +1810,7 @@ if (typeof submitBundleFile == 'undefined') {
     const submitBundleFile = "SubmitButton.jsx";
 }
 
-// Validate that the RenderQueueIndex for each selectionItem is still valid
+// Validate that the RenderQueueIndex for each selectionItem is still valid and update list if they're out of date
 function UpdateRenderQueueIndices(renderQueueIndex, selectionItem) {
     if (
         renderQueueIndex < 1 ||
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -2131,9 +2131,9 @@ function SubmitSelection(selection, selectionSettings, framesPerTask, multiFrame
         logger.debug("OutputFolder is: " + outputFolder, submitBundleFile);
 
         // Calculate frame range using the utility function
-        const frameRange = dcUtil.calculateFrameRange(renderQueueItem);
-        const startFrame = frameRange.startFrame;
-        const endFrame = frameRange.endFrame;
+        var frameRange = dcUtil.calculateFrameRange(renderQueueItem);
+        var startFrame = frameRange.startFrame;
+        var endFrame = frameRange.endFrame;
 
         var dependencies = findJobAttachments(renderQueueItem.comp); // list of filenames
         var compName = dcUtil.removeIllegalCharacters(renderQueueItem.comp.name);
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -3196,8 +3196,8 @@ function buildUI(thisObj) {
             columnTitles: ['#', 'Name', 'Frames', 'Output Path'],
             columnWidths: [32, 160, 120, 240],
         });
-        newList.preferredSize.height = 400
-        newList.preferredSize.width = 500
+        newList.preferredSize.height = 200;
+        newList.preferredSize.width = 500;
 
         for (var i = 1; i <= app.project.renderQueue.numItems; i++) {
             var rqi = app.project.renderQueue.item(i);
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1944,7 +1944,7 @@ function generateStepParameterFragment(bundlePath, isImageSeq, renderQueueIndex,
 }
 
 // Generates the step chunk of the template for each step by loading the `step_<>_fragment.json`
-//      Replacing the parmaeters to be pointing to our per-renderQueueItem parameters and updating any parameters in the onRun
+//      Replacing the parameters to be pointing to our per-renderQueueItem parameters and updating any parameters in the onRun
 function generateStepTemplateFragment(bundlePath, isImageSeq, renderQueueItemIndex, compName, taskTimeoutSeconds) {
     var path = bundlePath + "/step_video_fragment.json";
     if (isImageSeq) {
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -3243,11 +3243,6 @@ function buildUI(thisObj) {
     const submitButton = controlsGroup.add("button", undefined, "Submit");
     submitButton.onClick = function() {
         if (getPythonExecutable()) {
-            const multiFrameRendering = mfrCheckBox.value ? "ON" : "OFF";
-            var maxCpuUsagePercentage = undefined;
-            if (mfrCheckBox.value) {
-                maxCpuUsagePercentage = parseInt(maxCpuUsagePercentageTextBox.text);
-            }
             if (list.selection === null) {
                 return;
             }
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1054,8 +1054,11 @@ function UiSettingsState() {
 }
 function UiSettingsStore(name) {
     this.name = name;
+    // _framesPerTask: string
     this._framesPerTask = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK);;
+    // _multiFrameRendering: bool
     this._multiFrameRendering = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING);
+    // _maxCpuUsagePercentage: string
     this._maxCpuUsagePercentage = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE);
 
     this.framesPerTask = function () {
@@ -1063,7 +1066,7 @@ function UiSettingsStore(name) {
     }
     this.setFramesPerTask = function (value) {
         logger.warning("(" + this.name + ") Setting framesPerTask to " + value)
-        this._framesPerTask = value
+        this._framesPerTask = typeof value === "string" ? value : value.toString()
     }
 
     this.multiFrameRendering = function () {
@@ -1071,7 +1074,7 @@ function UiSettingsStore(name) {
     }
     this.setMultiFrameRendering = function (value) {
         logger.warning("(" + this.name + ") Setting multiFrameRendering to " + value)
-        this._multiFrameRendering = value
+        this._multiFrameRendering = typeof value === "boolean" ? value : (value === "true")
     }
 
     this.maxCpuUsagePercentage = function () {
@@ -1079,8 +1082,26 @@ function UiSettingsStore(name) {
     }
     this.setMaxCpuUsagePercentage = function (value) {
         logger.warning("(" + this.name + ") Setting maxCpuUsagePercentage to " + value)
-        this._maxCpuUsagePercentage = value
+        this._maxCpuUsagePercentage = typeof value === "string" ? value : value.toString()
     }
+}
+
+UiSettingsState.prototype.create = function (compId, framesPerTask, multiFrameRendering, maxCpuUsagePercentage) {
+    if (!this.settings[compId]) {
+        this.settings[compId] = new UiSettingsStore(compId)
+    }
+    if (framesPerTask === undefined) {
+        framesPerTask = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK);
+    }
+    if (multiFrameRendering === undefined) {
+        multiFrameRendering = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING);
+    }
+    if (maxCpuUsagePercentage === undefined) {
+        maxCpuUsagePercentage = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE);
+    }
+    this.settings[compId].setFramesPerTask(framesPerTask);
+    this.settings[compId].setMultiFrameRendering(multiFrameRendering);
+    this.settings[compId].setMaxCpuUsagePercentage(maxCpuUsagePercentage);
 }
 
 UiSettingsState.prototype.get = function(compId) {
@@ -1127,7 +1148,7 @@ function parameterValues(
         },
         {
             name: prefix + "_MultiFrameRendering",
-            value: multiFrameRendering,
+            value: multiFrameRendering === true ? "ON" : "OFF",
         },
     ];
     if (maxCpuUsagePercentage) {
@@ -2686,7 +2707,67 @@ function buildUI(thisObj) {
     const listGroup = root.add("panel", undefined, "");
     listGroup.alignment = ['fill', 'fill'];
     listGroup.alignChildren = ['fill', 'fill']
-    var list = null;
+
+    const bounds = list == null ? undefined : list.bounds;
+    var list = listGroup.add("listbox", bounds, "", {
+        multiselect: true,
+        numberOfColumns: 4,
+        showHeaders: true,
+        columnTitles: ['#', 'Name', 'Frames', 'Output Path'],
+        columnWidths: [32, 160, 120, 240],
+    });
+    list.preferredSize.height = 400;
+    list.preferredSize.width = 500;
+
+    function onSelectionChange() {
+        const selection = list.selection;
+        if (selection == null) {
+            refreshList(list, uiSettingsState);
+            framesPerTaskTextBox.text = "";
+            return;
+        }
+        submitButton.enabled = true;
+        submitButton.active = false;
+        submitButton.active = true;
+
+        // Disable everything
+        framesPerTaskTextBox.enabled = false
+        mfrCheckBox.enabled = false
+        maxCpuUsagePercentageTextBox.enabled = false
+
+        if (selection.length !== 1) {
+            return
+        }
+        const selectionItem = selection[0]
+        logger.warning("Selected Comp is: " + app.project.renderQueue.item(selectionItem.renderQueueIndex).comp.name);
+        const imageOutput = isRenderQueueItemImageOutput(app.project.renderQueue.item(selectionItem.renderQueueIndex))
+        framesPerTaskTextBox.enabled = imageOutput
+        mfrCheckBox.enabled = true
+        maxCpuUsagePercentageTextBox.enabled = true
+
+        logger.debug("    Setting framesPerTaskTextBox.text to: " + selectionItem.subItems[1].text);
+        framesPerTaskTextBox.text = selectionItem.subItems[1].text
+
+        const settings = uiSettingsState.get(selectionItem.compId)
+        if (settings === undefined) {
+            logger.warning("Could not find settings for : " + selectionItem.compId);
+            return
+        }
+
+        if (imageOutput === true) {
+            logger.debug("    Setting framesPerTaskTextBox.text to: " + (settings.framesPerTask() || selectionItem.subItems[1].text));
+            framesPerTaskTextBox.text = settings.framesPerTask() || selectionItem.subItems[1].text
+        }
+        logger.debug("    Setting mfrCheckBox.value to: " + settings.multiFrameRendering());
+        mfrCheckBox.value = settings.multiFrameRendering()
+        logger.debug("    Setting maxCpuUsagePercentageTextBox.text to: " + settings.maxCpuUsagePercentage());
+        maxCpuUsagePercentageTextBox.text = settings.maxCpuUsagePercentage()
+
+        maxCpuUsagePercentageTextBox.enabled = mfrCheckBox.value
+    }
+
+    list.onChange = onSelectionChange;
+
     const controlsGroup = root.add("group", undefined, "");
     controlsGroup.orientation = 'column';
     controlsGroup.alignment = ['fill', 'bottom'];
@@ -2727,6 +2808,9 @@ function buildUI(thisObj) {
         }
         app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK, framesPerTaskTextBox.text);
         for (var s = 0; s < list.selection.length; s++) {
+            if (list.selection == null) {
+                return;
+            }
             const selectionItem = list.selection[s];
             uiSettingsState.get(selectionItem.compId).setFramesPerTask(framesPerTaskTextBox.text)
         }
@@ -3016,12 +3100,15 @@ function buildUI(thisObj) {
     }
 
     updateList();
+    refreshList(list, uiSettingsState);
     if (list.selection != null && list.selection.length === 1) {
         const selectionItem = list.selection[0]
         const renderQueueItem = app.project.renderQueue.item(selectionItem.renderQueueIndex)
         framesPerTaskTextBox.enabled = isRenderQueueItemImageOutput(renderQueueItem)
     }
-    refreshButton.onClick = updateList;
+    refreshButton.onClick = function() {
+        refreshList(list, uiSettingsState);
+    }
 
     submitterPanel.layout.layout(true);
 
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1874,8 +1874,38 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
     const startFrame = frameRange.startFrame;
     const endFrame = frameRange.endFrame;
 
-    const aftereffectsVersion = app.version[0] + app.version[1];
-    logger.debug("The major version of After Effects is " + aftereffectsVersion, submitBundleFile);
+    // Check if warning should be shown
+    const ignoreWarning = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING) === "true";
+    const savedVersion = parseFloat(app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING_VERSION) || "0");
+    const currentVersion = dcUtil.getAEVersion();
+
+    // Is this AE version not supported in the deadline-cloud channel?
+    if (SUPPORTED_VERSIONS.indexOf(currentVersion) === -1) {
+        // If so, has the warning already been ignored or is the user on a different AE version and we should warn them again?
+        if (!ignoreWarning || savedVersion !== currentVersion) {
+            const versionMismatchWarningMessage = "Warning: Your After Effects version " + currentVersion +
+            " is not officially supported in the deadline-cloud conda channel. Supported versions are: " + SUPPORTED_VERSIONS.join(", ") + ". " +
+            "This may result in compatibility issues or failed jobs.\n\nDon't show this warning again for version " + currentVersion + "?";
+
+            // Provide warning, and if acknowledged, store their current version and warning preference. Otherwise, block job submission.
+            app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING_VERSION, currentVersion.toString());
+            if (confirm(versionMismatchWarningMessage)) {
+                app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING, "true");
+            } else {
+                app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING, "false");
+                return;
+            }
+        } else {
+            logger.debug("Version mismatch already acknowledged, version warning skipped.");
+        }
+        logger.debug("Defaulting to After Effects major version conda package to minimize incompatibility issues.");
+    }
+
+    var aftereffectsCondaVersion = dcUtil.getAEVersion();
+    if (SUPPORTED_VERSIONS.indexOf(aftereffectsCondaVersion) === -1) {
+        aftereffectsCondaVersion = Math.floor(aftereffectsCondaVersion);
+    }
+    logger.debug("The compatible version of After Effects is " + aftereffectsCondaVersion, submitBundleFile);
 
     /**
      * Generates parameter_values json file
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -926,9 +926,8 @@ function __generateUtil() {
          * @returns {Object} Object containing startFrame and endFrame
          */
         // NOTE: we're not using displayStartFrame since it is rounded up
-        const startFrame = Number(Math.floor(rqi.timeSpanStart * rqi.comp.frameRate));
-        const numFrames = Number(Math.floor(rqi.timeSpanDuration * rqi.comp.frameRate));
-        const endFrame = startFrame + numFrames - 1; // end frame is inclusive
+        const startFrame = Number(Math.round(rqi.timeSpanStart * rqi.comp.frameRate));
+        const endFrame = Number(Math.round((rqi.timeSpanStart + rqi.timeSpanDuration) * rqi.comp.frameRate))
         return {
             startFrame: startFrame,
             endFrame: endFrame
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1050,6 +1050,49 @@ var logger = Logger(logFileName, logNormDirectoryPath);
 
 
 
+function UiSettingsState() {
+    this.settings = {}
+}
+function UiSettingsStore(name) {
+    this.name = name;
+    this._framesPerTask = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK);;
+    this._multiFrameRendering = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING);
+    this._maxCpuUsagePercentage = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE);
+
+    this.framesPerTask = function () {
+        return this._framesPerTask
+    }
+    this.setFramesPerTask = function (value) {
+        logger.warning("(" + this.name + ") Setting framesPerTask to " + value)
+        this._framesPerTask = value
+    }
+
+    this.multiFrameRendering = function () {
+        return this._multiFrameRendering
+    }
+    this.setMultiFrameRendering = function (value) {
+        logger.warning("(" + this.name + ") Setting multiFrameRendering to " + value)
+        this._multiFrameRendering = value
+    }
+
+    this.maxCpuUsagePercentage = function () {
+        return this._maxCpuUsagePercentage
+    }
+    this.setMaxCpuUsagePercentage = function (value) {
+        logger.warning("(" + this.name + ") Setting maxCpuUsagePercentage to " + value)
+        this._maxCpuUsagePercentage = value
+    }
+}
+
+UiSettingsState.prototype.get = function(compId) {
+    if (!this.settings[compId]) {
+        this.settings[compId] = new UiSettingsStore(compId)
+    }
+    return this.settings[compId]
+}
+
+
+
 var jobTemplateHelperFile = "JobTemplateHelper.json";
 /**
  * Generates the basic parameterValue file for the job template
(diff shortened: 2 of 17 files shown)


Ahuge:
Force pushed just to make DCO signoff happy

Ahuge:
Pushing the DCO signoff again, sorry

src/utils/Logger.jsx
```diff
@@ -83,8 +83,8 @@ function Logger(logFileName, logDirectoryPath, maxBytes, backupCount) {
             if (!rolloverFile.exists) {
                 continue;
             }
-            var j = i + 1;
```
eherozhao, line 86: I think var here makes more sense than const.

src/utils/Utils.jsx
```diff
@@ -1,17 +1,17 @@
 var scriptFolder = Folder.current.fsName;
-const SUPPORTED_VERSIONS = [24.6, 25.1, 25.2];
+var SUPPORTED_VERSIONS = [24.6, 25.1, 25.2];
```
eherozhao: Should this be a constant variable?

Ahuge:
Pushing the DCO signoff again because I keep forgetting to DCO signoff.

Sorry

src/utils/CompositionSettings.jsx
```diff
@@ -0,0 +1,61 @@
+function UiSettingsState() {
+    this.settings = {}
+}
+function UiSettingsStore(name) {
+    this.name = name;
+    // _framesPerTask: string
+    this._framesPerTask = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK);;
```
eherozhao: two `;`

eherozhao (commented):
Thanks for raising this feature up!!

1. Can we add some screenshot about how it will look?
2. Can you confirm this will not break existing functionality like, image chunking, video/image rendering?
3. We can add README introducing this feature next once this merge.

Ahuge:
Here is the image of the feature:
image

And an interactive GIF of it being used
deadline-cloud-multi-submit

Ahuge:
- Can you confirm this will not break existing functionality like, image chunking, video/image rendering?
  | Can confirm that I have used this with the image chunking and tested both image and video outputs.

src/ui/SubmitterUI.jsx
```diff
@@ -99,7 +215,15 @@ function buildUI(thisObj) {
             framesPerTaskTextBox.text = newFramesPerTaskValue;
         }
         app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK, framesPerTaskTextBox.text);
+        if (list.selection == null) {
```
viknith, line 72: To make sure I understand this correctly, when this setting is modified, it will apply to all compositions that are selected? And this is the case for other settings as well?

src/submission/SubmitBundle.jsx
```diff
@@ -22,200 +21,427 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+        outputFileName,
+        isImageSeq,
+        startFrame,
+        endFrame,
+        chunkSize,
+        multiFrameRendering,
+        maxCpuUsagePercentage,
+        prefix,
+    )
+}
```
viknith: At a first glance, this function seems redundant, can't we just use rename parameterValue() to generateParameterValuesForStep() or some variation of that?

src/submission/SubmitBundle.jsx
```diff
@@ -1,19 +1,18 @@
...
-    // first we must verify that our selection is valid
-    if (selection == null) {
-        adcAlert("Error: No selection", true);
-        return;
-    }
+var JobParams = [
+    "JobScriptDir",
+    "CondaPackages",
+    "ProjectFile"
+]
```
viknith, line 5: If we're going to match in all places where `Param.` is used, do we need this JobParams list at all? There appears to be potential redundancy.

viknith (commented):
To make sure I understand this correctly at a high-level:
- We are making parameters like `comp1_ProjectFile`, `comp1_RenderQueueIndex`, `comp2_ProjectFile`, `comp2_RenderQueueIndex`, etc.
- For each composition, a step is created and a step fragment is used to generate that step and override any missing values.
- All the JSONs for parameter definition fragments is aggregated into one parameter defn file and all the JSONs for step definition fragments is aggregated into one file.

src/ui/SubmitterUI.jsx
Ahuge: That is correct, but these UI elements are only enabled with one selection. So this is actually a slightly safer way of doing list.selection[0]

There *should* be no difference in practice between those.

I can swap it to the clearer implementation of you'd like

src/submission/SubmitBundle.jsx
Ahuge: Good catch, I'll remove this function and just call it raw

src/submission/SubmitBundle.jsx
Ahuge: We're building our regex based off of these "Non-Step Parameters". I could inline them in the regex but I felt this was clearer.

Let me know your thoughts

Ahuge:
> To make sure I understand this correctly at a high-level:
> - We are making parameters like `comp1_ProjectFile`, `comp1_RenderQueueIndex`, `comp2_ProjectFile`, `comp2_RenderQueueIndex`, etc.
> - For each composition, a step is created and a step fragment is used to generate that step and override any missing values.
> - All the JSONs for parameter definition fragments is aggregated into one parameter defn file and all the JSONs for step definition fragments is aggregated into one file.

That is correct.
The only other note is the Job environment where we create output files gets modified to take each steps' output dir as well

viknith:
Here's a video where job submission failed silently

https://github.com/user-attachments/assets/593772ad-8506-478a-a4ed-70b9db984280

I also found that the MFR settings were not persisting between compositions and stayed the same. Didn't we want them to be different on a per composition basis? Video for that:

https://github.com/user-attachments/assets/c2298b3f-e3b9-4d0a-ade0-242a04074165

viknith:
A few points/nits:
- On Mac, the multi-select isn't Control-Click, it's Command-Click, so that wording needs to be updated based on OS type
- I noticed the panel is really really long now and I can only see the whole panel if I stretch it to the max in order to click the submit button, which required undocking the submitter. Can the composition list box be shorter to fix that?
- precomp -> compositions, keep it consistent

viknith:
Just tried submitting a composition, here's an error that the GUI submitter popped up

Screenshot 2025-07-30 at 2 49 11 PM

Seems like the labels are getting too long in the OpenJD template and we aren't handling that smoothly. Here's a doc on our OpenJobDescription (OpenJD or OJD) templates and the limitations on each field: https://github.com/OpenJobDescription/openjd-specifications/blob/c5ab3ad69cc21786e6b58d66a9c09f31f248daf5/wiki/2023-09-Template-Schemas.md

src/submission/SubmitBundle.jsx
viknith: Is there a way to dynamically generate this list in case we ever added non-step parameters in the future so that we don't miss that here? Or a mechanism we can put in place so that future developers won't miss updating this field?

src/submission/SubmitBundle.jsx
```diff
@@ -1,28 +1,21 @@
...
-    var renderQueueIndex = selection.renderQueueIndex;
-    var rqi;
+if (typeof submitBundleFile == 'undefined') {
+    const submitBundleFile = "SubmitButton.jsx";
+}
 
-    // because our panel is updated independently of the render queue, the two may become out of sync
-    // we need to verify that the selection made actually matches what is in the render queue
+// Validate that the RenderQueueIndex for each selectionItem is still valid
+function UpdateRenderQueueIndices(renderQueueIndex, selectionItem) {
```
viknith, line 18: Nit: rename from `UpdateRenderQueueIndices` if the purpose of this function is for validation only. Update in the function name makes it odd.

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,242 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+    const outputModule = renderQueueItem.outputModule(1).file;
+    if (outputModule == null) {
+        adcAlert("Error: Render Queue Item " + renderQueueItem.comp.name + " does not have its output file set", true);
+        return false;
+    }
+    return true;
+}
+
+// Generate our prefixed Parameter Values for the provided comp
+function generateParameterValuesForStep(
```
viknith: Is this function redundant or is there a need to wrap the parameterValues function? Could we just rename the parameterValues function to be generateParameterValuesForStep?

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,242 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+    const updatedParameterDefinitions = []
+    for (var i = 0; i < stepParametersObject.parameterDefinitions.length; i++) {
+        if (JobParams.indexOf(stepParametersObject.parameterDefinitions[i].name) !== -1) {
+            // Don't modify these values
+            continue;
+        }
+        var replacedDefinition = stepParametersObject.parameterDefinitions[i]
+        replacedDefinition.name = renderQueueItemID + "_" + stepParametersObject.parameterDefinitions[i].name;
+        replacedDefinition.userInterface.label = "(" + renderQueueItemID + ") " + replacedDefinition.userInterface.label;
+
```
viknith: Spoke about this offline, we'll need the labels to be shorter. But additionally we should add some error handling here to surface to the customer that there's an issue. We could recommend that they shorten the composition name to a specific range.

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,242 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+    stepTemplateObject.steps[0].name = renderQueueItemID;
+    // Replace any parameter names in onRun script
+    const scriptArgs = stepTemplateObject.steps[0].script.actions.onRun.args;
+    const replacedArgs = []
+    for (var i = 0; i < scriptArgs.length; i++) {
+        // JobParams
+        replacedArgs.push(scriptArgs[i].replace(paramPatternRegex, "Param." + renderQueueItemID + "_"));
+    }
+    stepTemplateObject.steps[0].script.actions.onRun.args = replacedArgs;
+    if (stepTemplateObject.steps[0].script.actions.onRun) {
```
viknith: Nit: would onRun ever not be defined?

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,242 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+        replacedDefinition.userInterface.label = "(" + renderQueueItemID + ") " + replacedDefinition.userInterface.label;
+
+        updatedParameterDefinitions.push(replacedDefinition);
+    }
+    stepParametersObject.parameterDefinitions = updatedParameterDefinitions;
+    return stepParametersObject;
+}
+
+// Generates the step chunk of the template for each step by loading the `step_<>_fragment.json`
+//      Replacing the parmaeters to be pointing to our per-renderQueueItem parameters and updating any parameters in the onRun
```
viknith: Nit: parameters instead of parmaeters

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,242 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+/**
+ * Submit the selected render queue item
+ **/
+function SubmitSelection(selection, selectionSettings) {
+    // Calculate task run timeout in seconds
+    var taskTimeoutSeconds = (selectionSettings.taskRunDays() * 24 * 60 * 60) + (selectionSettings.taskRunHours() * 60 * 60) + (selectionSettings.taskRunMinutes() * 60);
+    // Validate timeout values during job submission
+    if (taskTimeoutSeconds <= 0) {
+        adcAlert("The following timeout value must be greater than 0: TaskRun", true);
+        throw new Error("Task run timeout must be greater than zero");
```
viknith: Should a whole Error be thrown? Or could we just return and stop the the execution from continuing further? Asking because it doesn't appear like anything is catching the Error it would be simpler to return I think.

src/submission/SubmitBundle.jsx
```diff
@@ -88,177 +268,211 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+                jobParameterValues.parameterValues.push(parameterValues.parameterValues[p]);
+            }
         }
-        recursiveCopy(jobTemplateSourceFolder, bundleRoot);
 
-        generateTemplate(bundlePath, isImageSeq);
-        return bundleRoot;
+        stepOutputFolderParameters.push("{{Param." + renderQueueItemID + "_OutputDir}}");
+
+        var stepTemplate = generateStepTemplateFragment(bundle.fsName, isImageSeq, renderQueueItemID, taskTimeoutSeconds);
```
viknith: Could use some more commenting on this whole section of the SubmitSelection workflow after validation is done to help understand what's going on

src/ui/SubmitterUI.jsx
```diff
@@ -2,82 +2,100 @@
...
     const controlsPanel = controlsGroup.add("panel", undefined, "");
     controlsPanel.alignment = ['fill', 'top'];
 
-    // Container with all settings to modify job submission
-    const settingsGroup = controlsPanel.add("group", undefined, "");
-    settingsGroup.orientation = "column";
-    settingsGroup.alignment = ['fill', 'top'];
-    settingsGroup.alignChildren = ['left', 'top'];
+    // Container with settings to modify comp-specific settings
+    const perCompSettingsGroup = controlsPanel.add("panel", undefined, "Precomp-Specific Settings");
```
viknith: Nit: Percomp, not Precomp

src/ui/SubmitterUI.jsx
viknith: Implementation has been updated so that each comp has it's own percomp settings and also a section for global settings. Resolving

src/ui/SubmitterUI.jsx
```diff
@@ -92,156 +110,173 @@ function buildUI(thisObj) {
...
         const isMfrChecked = mfrCheckBox.value;
-        if (!isMfrChecked) {
-            maxCpuUsagePercentageTextBox.text = "N/A";
-            app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING, "false");
-        } else {
-            maxCpuUsagePercentageTextBox.text = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE);
-            app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING, "true");
+        const selectionItem = dcUtil.getSelection(list);
+        if (selectionItem) {
+            var RQIID = dcUtil.getRQIID(selectionItem.renderQueueIndex);
```
viknith: Nit: rename getRQIID to getRenderQueueIndexId

src/ui/SubmitterUI.jsx
```diff
@@ -92,156 +110,173 @@ function buildUI(thisObj) {
...
-        if (selection == null) {
-            return false;
+        taskRunDaysInput.text = uiSettingsState.taskRunDays();
+    }
+    taskRunDaysInput.onChange = onTaskRunDaysChanged;
+
+    function onTaskRunHoursChanged() {
+        var original_value = uiSettingsState.taskRunHours();
+        taskRunHoursInput.text = taskRunHoursInput.text.replace(/[^0-9]/g, "");
+        var new_value = parseInt(taskRunHoursInput.text);
```
viknith: Nit, newValue, not new_value to follow naming convention. Same for other spots except in the case of all-caps constants.

src/ui/SubmitterUI.jsx
```diff
@@ -92,156 +110,173 @@ function buildUI(thisObj) {
...
+    }
+    taskRunHoursInput.onChange = onTaskRunHoursChanged;
+
+    function onTaskRunMinutesChanged() {
+        var original_value = uiSettingsState.taskRunMinutes();
+        taskRunMinutesInput.text = taskRunMinutesInput.text.replace(/[^0-9]/g, "");
+        var new_value = parseInt(taskRunMinutesInput.text);
+        if (taskRunMinutesInput.text === "") new_value = 0;
+        if (dcUtil.validateTimeoutValues(taskRunCheckbox.value, taskRunDaysInput.text, taskRunHoursInput.text, taskRunMinutesInput.text)) {
+            if (!isNaN(new_value)) {
```
viknith: Silly question, why would the value be not a number if it passed the validateTimeoutValues() function

src/ui/SubmitterUI.jsx
```diff
@@ -260,11 +295,15 @@ function buildUI(thisObj) {
             var item = newList.add('item', i.toString());
             item.renderQueueIndex = i;
             item.compId = rqi.comp.id;
+            // Create a default entry for each comp as needed.
+            uiSettingsState.get(i);
```
viknith, line 305: If the comp on the 4th index is removed and the uiSettingsState is updated to just include the first 3, does the 4th one get evicted? More simply, how do we manage uiSettingsState as we add and remove compositions to ensure old settings don't persist for old compositions from the past?

src/utils/Utils.jsx
```diff
@@ -399,17 +482,17 @@ function __generateUtil() {
...
         // Wrapped command with error code output
         cmd = cmd + " > " + tempOutputFile.fsName;
-        cmd += "\nIF %ERRORLEVEL% NEQ 0 ("
-        cmd += "\n echo ERROR CODE: %ERRORLEVEL% >> " + tempOutputFile.fsName
-        cmd += "\n)"
-        cmd += "\nexit"
+        cmd += "\nIF %ERRORLEVEL% NEQ 0 (";
+        cmd += "\n echo ERROR CODE: %ERRORLEVEL% >> " + tempOutputFile.fsName;
+        cmd += "\n)";
+        cmd += "\nexit";
```
viknith, line 584: Were these originally causing issues without semicolons? Otherwise, is there any reason to add these?

src/utils/Utils.jsx
```diff
@@ -840,30 +954,31 @@ if (typeof DEADLINECLOUD_TASK_RUN_TIMEOUT_HOURS === "undefined") {
...
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_TASK_RUN_TIMEOUT_ENABLED)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_TASK_RUN_TIMEOUT_ENABLED, "10");
+if (typeof DEFAULT_MULTI_FRAME_RENDERING === "undefined") {
+    const DEFAULT_MULTI_FRAME_RENDERING = false;
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_TASK_RUN_TIMEOUT_DAYS)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_TASK_RUN_TIMEOUT_DAYS, "2");
+if (typeof DEFAULT_MAX_CPU_USAGE_PERCENTAGE === "undefined") {
+    const DEFAULT_MAX_CPU_USAGE_PERCENTAGE = 90;
```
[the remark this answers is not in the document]
viknith: Ignore

src/utils/Utils.jsx
```diff
@@ -840,30 +954,31 @@ if (typeof DEADLINECLOUD_TASK_RUN_TIMEOUT_HOURS === "undefined") {
...
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE, "90");
+if (typeof DEFAULT_FRAMESPERTASK === "undefined") {
+    const DEFAULT_FRAMESPERTASK = 10;
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_TASK_RUN_TIMEOUT_ENABLED)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_TASK_RUN_TIMEOUT_ENABLED, "10");
+if (typeof DEFAULT_MULTI_FRAME_RENDERING === "undefined") {
+    const DEFAULT_MULTI_FRAME_RENDERING = false;
```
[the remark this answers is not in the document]
viknith: Ignore

src/utils/Utils.jsx
```diff
@@ -840,30 +954,31 @@ if (typeof DEADLINECLOUD_TASK_RUN_TIMEOUT_HOURS === "undefined") {
...
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING, "false");
+if (typeof DEFAULT_TASK_RUN_TIMEOUT_MINUTES === "undefined") {
+    const DEFAULT_TASK_RUN_TIMEOUT_MINUTES = 0;
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MAX_CPU_USAGE_PERCENTAGE, "90");
+if (typeof DEFAULT_FRAMESPERTASK === "undefined") {
+    const DEFAULT_FRAMESPERTASK = 10;
```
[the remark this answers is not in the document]
viknith: Ignore

src/utils/Utils.jsx
```diff
@@ -840,30 +954,31 @@ if (typeof DEADLINECLOUD_TASK_RUN_TIMEOUT_HOURS === "undefined") {
...
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_FRAMESPERTASK, "10");
+if (typeof DEFAULT_TASK_RUN_TIMEOUT_HOURS === "undefined") {
+    const DEFAULT_TASK_RUN_TIMEOUT_HOURS = 0;
 }
-if (!app.settings.haveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING)) {
-    app.settings.saveSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_MULTI_FRAME_RENDERING, "false");
+if (typeof DEFAULT_TASK_RUN_TIMEOUT_MINUTES === "undefined") {
+    const DEFAULT_TASK_RUN_TIMEOUT_MINUTES = 0;
```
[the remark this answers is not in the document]
viknith: Ignore

src/utils/Utils.jsx
olly-glenn-arch: I thought I needed to add them in order to pass the Github checkers, but I can revert them if that wasn't necessary

src/ui/SubmitterUI.jsx
olly-glenn-arch: Right now we don't, but it would probably be pretty simple to implement this by checking for and removing unused settings each time the list updates. Right now persistence isn't perfect because I wasn't able to find a way to consistently and uniquely identify render queue items--they don't have any sort of ID natively as far as I'm aware, while their indices are unique they change if they're reordered, and comp names aren't unique because it's possible to submit multiple render queue items from the same comp. Project + index + comp name seems to make the most sense to use as an identifier right now because it seems like the simplest way to reduce the likelihood of collisions, but that means that if the comps are reordered and their indices change settings will get reset.
One thing that could be useful to look in to in the future is seeing if it's possible to store values in the render queue item's metadata instead of needing to store them in the global settings object, or alternatively figuring out a better way to consistently identify render queue items.

src/ui/SubmitterUI.jsx
olly-glenn-arch: Most of the timeout logic is stuff I inherited from the mainline rebase so I didn't write it initially, but I think the idea is that because the validateTimeoutValues method will only ever be called to validate the state of the UI, it's easier to pass in text values and let the validation function parse them then need to parse them individually before passing them to the function.

src/ui/SubmitterUI.jsx
olly-glenn-arch: Whoops, thought I caught all of those. Every once in a while my brain switched tracks into python mode and started using snake case, I'll switch everything over.

src/ui/SubmitterUI.jsx
olly-glenn-arch: I was actually thinking of changing the language to be "render queue item-specific settings" or "queue item-specific settings" or "item-specific-settings" or something similar because the settings don't actually apply to the comp, but rather the entry in the AE render queue.

src/submission/SubmitBundle.jsx
olly-glenn-arch: I agree, and the alert is probably enough to let the user know something has gone wrong.

olly-glenn-arch:
### Design Summary
The biggest architectural change in this PR is that a new object called `UiSettingsState` is now used to manage all of the values the user can control from the UI in a centralized way. This makes it easier to manage the increased number of parameters that the user can control now that they can set settings for individual comps.

When values are accessed from the UiSettingsState, they are read/written from the Aftereffects settings using a set of new utility methods that allow for the accessing of settings in a more type-safe way (`getBoolSetting`, `saveBoolSetting`, etc). All direct accesses to the Aftereffects settings have been replaced by these utility methods. The `get\<type\>` setting methods allow for the specification of default values and automatically create settings if they don't exist.

The default values for render job parameters are no longer saved to the settings at the end of `Utils.jsx` but instead are saved as constants that are used as default values when a `UiSettingsState` is created. 

The `UiSettingsState` object contains a dictionary of `UiSettingsStore` objects that are used to store settings unique to each render queue item. These items are prefixed by their IDs when saved to the Aftereffects settings object so that they can be identified. The way IDs are generated for render queue items is defined in the `getRQIID` (soon to be renamed to `getRenderQueueItemID` method--currently it is `\<project name\>_\<render_queue_index\>_\<comp_id\>`. Unfortunately Aftereffects does not natively supply a way to get unique IDs for render queue items and implementing them ourselves is a nontrivial task, so the goal here is to reduce the likelihood of collisions as much as possible, not to create IDs that are globally unique.

When render jobs are submitted, each render queue item is submitted as a step in the job. The submission JSON is constructed by starting with the default `template.json` file, then using `parameter_definitions_video_fragment.json` and `parameter_definitions_image_fragment.json` to construct the parameter definitions for each render queue item, then using `step_video_fragment.json` and `step_image_fragment.json` to construct each step of the render job by pointing them at the parameter definitions, and finally putting all the fragments together and writing everything out to the bundle. The names of the parameters for each render queue item are also prefixed by an ID, but the ID used here is different then the one used in the UiSettingsStore. The ID used for render queue items in job submission is `index_compName_parameterName`. Compared to the ID used in the settingsStore, this one has the project name removed because all render jobs can be assumed to come from the same project, and uses compName instead of compID so that users will be able to better understand which render queue item is being referenced. Because the index guarantees uniqueness, the compName can be safely truncated to keep the parameter name under the 64 character limit for parameter names in job bundles.

This PR also contains tweaks to the UI layout to make it more clear which settings are local to render queue items and which settings are global to the entire job, as well as disabling items when they are not relevant to the current selection, adding language to the UI to make it more clear that multi-comp selection is an option, and other small bug fixes relating to keeping the data in the UI synchronized with the stored settings.

### Pending Changes
- Remove unused data store values when list of render queue items changes
- Replace snake case variable names with camel case ones
- Rename getRQIID to getRenderQueueIndexId for clarity
- Truncate comp names when used in job parameters to stay below character limit
- Update language to refer to items as render queue items, not comps
- Miscellaneous typos and tweaks from review

viknith:
I've tested the PR up to this point and job submission is working AE 25 and 24 on Mac!

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1864,9 +1864,40 @@ function generateBundle() {
...
+    /** Generates the name of the parameter to be used in the job submission
+     * by prefixing with the index and compName
+     * Truncates compname to avoid going over character limit.
+     **/
+    if (parameter === undefined) {
+        parameter = "";
+    }
+    if (parameter !== "") {
+        parameter = "_" + parameter;
+    }
```
viknith, line 1900: Shouldn't this be if `parameter === ""`?

olly-glenn-arch: That was intentional, we only want to prefix the parameter with an underscore if a parameter is specified. If the parameter field is blank then we don't want to add an underscore because that would just add a trailing underscore to the end of the resulting string

src/ui/SubmitterUI.jsx
olly-glenn-arch: It looks like it's actually not possible to remove settings from the `app.settings` object--you can read and write settings, but not delete them ([https://ae-scripting.docsforadobe.dev/other/settings/?h=](https://ae-scripting.docsforadobe.dev/other/settings/?h=)). It would probably be a good idea to move to a different method of storing render settings in the future, but I don't think it's worth it to change for this PR.

src/submission/SubmitBundle.jsx
olly-glenn-arch: This method actually does update the render queue indices, if they're stale it calls updateList(). Updated the comment to make it more clear

src/submission/SubmitBundle.jsx
olly-glenn-arch: Just consolidated into one generateParameterValues method

src/submission/SubmitBundle.jsx
olly-glenn-arch: Added logic that truncates comp names to keep everything under the character limit

src/submission/SubmitBundle.jsx
olly-glenn-arch: Yeah it looks like onRun is explicitly defined in both of the referenced template files so it will always be there, I removed the check.

src/submission/SubmitBundle.jsx
olly-glenn-arch: Fixed

src/submission/SubmitBundle.jsx
olly-glenn-arch: Added

src/ui/SubmitterUI.jsx
olly-glenn-arch: Done

src/ui/SubmitterUI.jsx
olly-glenn-arch: Ended up switching us over to XMPMetadata due to some other restrictions we found on storing things in the aftereffects settings

olly-glenn-arch:
### Updated Design Summary
While I think multicomp is a good name for this PR because it's clear and concise, this PR actually allows for the submission of multiple render queue items. A comp is a timeline in the project that contains visual data, and the render queue item is a specific render job associated with a comp that specifies codec information, output settings, and other render parameters. This distinction is important because it is possible to create multiple render queue items from the same comp--for example, you could create one render queue item that renders the comp to an image sequence and another that renders the same comp to an H264 video.

There are two large architectural changes in this PR. The first is that a new object called UiSettingsState is used to manage all of the values the user can control from the UI in a centralized way. This makes it easier to manage the increased number of parameters that the user can control now that they can set settings for individual comps. The second is that the plugin now stores values in the XMPMetadata embedded in the project file rather than the Aftereffects application settings. This fixes a variety of problems with saving render settings in the application settings that would be exacerbated by this PR, including:
- XMPMetadata is stored in the project file, so all changes are local to the Aftereffects project
    - This means that render settings will no longer be lost if the file is moved or the user folder is cleared, and it is impossible for render settings in one project to contaminate the render settings of another project
- It is not possible to remove application settings once you set them, but XMPMetadata values can be read, written, and deleted
- XMPMetadata is arranged in an XML-style tree structure that makes it easier to manage large quantities of parameters

When values are accessed from the UiSettingsState, they are read/written from the XMPMetadata using a set of new utility methods that allow for the accessing of settings in a more type-safe way (getBoolMetadata, saveBoolMetadata, etc). All direct accesses to the project settings have been replaced by these utility methods. They allow for the specification of default values and automatically create settings if they don't exist.

The default values for render job parameters are no longer saved to the settings at the end of Utils.jsx but instead are saved as constants that are used as default values when a UiSettingsState is created.

The UiSettingsState object contains a dictionary of UiSettingsStore objects that are used to store settings unique to each render queue item. These items are prefixed by their IDs when saved to the Aftereffects settings object so that they can be identified. The way IDs are generated for render queue items is defined in the getRenderQueueItemID method--currently it is _\<render_queue_index\>_\<comp_id\>. Unfortunately Aftereffects does not natively supply a way to get unique IDs for render queue items and implementing them ourselves is a nontrivial task, so the goal here is to reduce the likelihood of collisions as much as possible, not to create IDs that are globally unique. I think that finding a long-term robust solution for uniquely identifying render queue items is outside of the scope of this PR, but if we eventually do devise a better method then we'd just have to update the getRenderQueueItemID method and everything else will update.

When render jobs are submitted, each render queue item is submitted as a step in the job. The submission JSON is constructed by starting with the default template.json file, then using parameter_definitions_video_fragment.json and parameter_definitions_image_fragment.json to construct the parameter definitions for each render queue item, then using step_video_fragment.json and step_image_fragment.json to construct each step of the render job by pointing them at the parameter definitions, and finally putting all the fragments together and writing everything out to the bundle. The names of the parameters for each render queue item are also prefixed by an ID, but the ID used here is different then the one used in the UiSettingsStore. The ID used for render queue items in job submission is index_compName_parameterName. Compared to the ID used in the settingsStore, this one uses compName instead of compID so that users will be able to better understand which render queue item is being referenced. Because prefixing with the render queue index guarantees uniqueness, the compName can be safely truncated to keep the parameter name under the 64 character limit for parameter names in job bundles.

This PR also contains tweaks to the UI layout to make it more clear which settings are local to render queue items and which settings are global to the entire job, as well as disabling items when they are not relevant to the current selection, adding language to the UI to make it more clear that multi-comp selection is an option, and other small bug fixes relating to keeping the data in the UI synchronized with the stored settings.

olly-glenn-arch:
I just finished testing the multicomp PR on 25.1, 25.2, and 24.6 on windows and everything seems to be working correctly (except for importing image sequences, which is a known issue with a fix here https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/214). Unfortunately I don't have access to a mac, so if either someone else is available to test tomorrow or has a way I can access one that would be much appreciated.

jaikish (dismissed):

.gitignore
```diff
@@ -35,6 +35,7 @@ node_modules
 .idea
 *.code-workspace
 .vscode
+jsconfig.json
```
viknith, line 38: Non-blocking, we should document how to set this up in a later PR!

src/utils/CompositionSettings.jsx
```diff
@@ -0,0 +1,88 @@
+function UiSettingsState() {
```
viknith, line 1: Non-blocking, but should have a markdown file somewhere that explains all of this and the "dictionary structure" of how we're storing these parameters.

src/utils/Utils.jsx
```diff
@@ -107,9 +173,16 @@ function adcAlert(message, errorIcon) {
...
+    var actualType = typeof item;
+    if (actualType !== itemType) {
+        throw new Error("Object has type " + actualType + " instead of desired type " + itemType);
+    }
+}
+
 function __generateUtil() {
 
-    var scriptFileUtilName = "Util.jsx";
+    const scriptFileUtilName = "Util.jsx";
```
viknith: Utils.jsx, right?

src/utils/Utils.jsx
viknith: Ah gotcha nvm, missed that

src/utils/Utils.jsx
```diff
@@ -750,21 +921,90 @@ function __generateUtil() {
...
+        }
+        return true;
+    }
+
+    function getSelection(list) {
+        for (var s = 0; s < list.selection.length; s++) {
+            return list.selection[s];
+        }
+        return list.selection[0];
+    }
```
viknith: Does this just returning the first selection? If so why not just check the length and return the first one if the length is > 0.

src/utils/Utils.jsx
```diff
@@ -750,21 +921,90 @@ function __generateUtil() {
...
+        const regex = /.*xmp:(.*)/;
+        return regex.exec(path)[1];
+    }
+
+    function deleteUnusedMetadata(renderMetadataRoot) {
+        var metadata = new XMPMeta(app.project.xmpPacket);
+        var paths = []
+        var iterator = metadata.iterator(XMPConst.ITERATOR_JUST_CHILDREN, XMPConst.NS_XMP, renderMetadataRoot);
+        var property;
+        while (property = iterator.next()) {
```
viknith, line 975: Is this checking that iterator.next() is truthy and also assigns the value to the property variable?

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,243 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
         );
         updateList();
-        return;
+        return false;
     }
-    if (rqi.numOutputModules > 1) {
+    if (renderQueueItem.numOutputModules > 1) {
         adcAlert(
             "Warning: Multiple output modules detected. It is not supported in current submitter. Please raise an issue on Github repo for feature request.", false
         );
```
viknith, line 41: Should we call validateRenderQueueItemOutputModule() here as well?

src/submission/SubmitBundle.jsx
```diff
@@ -31,55 +24,243 @@ function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUs
...
+            // If the user hit yes to the prompt, but the file had never been saved, a second prompt would appear asking where they would want to save the project.
+            // If they hit cancel on the second prompt, the project file should be null and we should cancel the submission.
+            return;
+        }
+    }
+
     // Check if warning should be shown
-    const ignoreWarning = app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING) === "true";
-    const savedVersion = parseFloat(app.settings.getSetting(DEADLINECLOUD_SUBMITTER_SETTINGS, DEADLINECLOUD_IGNORE_VERSION_WARNING_VERSION) || "0");
+    const ignoreWarning = dcUtil.getBoolMetadata(dcUtil.composeXMPPath(DEADLINECLOUD_SETTINGS_ROOT, DEADLINECLOUD_IGNORE_VERSION_WARNING)) === "true";
```
viknith, line 247: Non-blocking, this gets a bool metadata but we're comparing it to a string called "true". Could we update the getBoolMetadata function to cast from string to boolean so that we're actually getting a boolean?

viknith:
Screenshot 2025-08-08 at 11 28 37 AM
Seems like the start frame is incorrect again, which was fixed in the most recent release. Seems like the most recent commits for frame range calculation is not working.

src/utils/Utils.jsx
Ahuge: https://extendscript.docsforadobe.dev/scripting-xmp/xmpscript-object-reference/?h=xmpmeta#xmpiteratornext

iterator.next() returns object or null

So we're just getting all values I think

viknith: Sounds like a yes, thank you!

viknith (dismissed):

jaikish (dismissed):

viknith (approved):

jaikish (approved):

Issue #29, opened 2024-05-03
Bug: Submitter reporting Error 9009 and is not able to find profile


yuanmich2:
### Expected Behaviour

When I start the submitter for the first time and click Login, I expect the submitter to be able to log me into the service and auto-populate my profile from the Deadline Monitor.

### Current Behaviour

The submitter shows green statuses at the bottom, but fails to get the profile or my farm and queue ids.
![image](https://github.com/aws-deadline/deadline-cloud-for-after-effects/assets/160265179/16cbb95f-fd14-49e1-9424-4b72a1a1e695)

Running "deadline config gui" in a command line and setting defaults for the profile, farm, and queue resolved my issue, but not everyone is going to know about that command. Additionally if I set the profile to "(default)" within the config gui, the submitter begins crashing.

### Reproduction Steps

Spin up new Windows Server 2019 EC2 instance. Install python 3.11, the aws cli, Deadline Cloud Monitor, and the latest After Effects. pip install deadline and deadline-cloud-for-after-effects. Clone this repo and install the submitter according to the readme. Create a new project and try to submit.

### Code Snippet

n/a

yuanmich2:
This is resolved in the newest version of the submitter that is coming out soon.

PR #212, opened 2025-06-26
feat: Ignore Missing Dependencies

### LINKED ISSUE #211: Feature request: Allow submitting a composition with missing dependencies
### Describe the problem

Part of a request to implement some D10 feature parity.

We would like the ability to submit a scene with missing asset references on the client machine.

### Proposed Solution

I think we could add a per-composition setting that allows us to skip local client validation of the existence of asset references in the scene and add the reference as a job dependency regardless.

### Example Use Cases

To be filled in

Ahuge:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/211

If you would like to submit your composition and ignore any warnings about missing dependencies, this was not currently possible. Currently the popup comes up letting you know that the file is missing but we wish to optionally silence this popup.

In addition, this PR passes `-continueOnMissingFootage` on the worker side if you have ignored missing footage.
Build upon the work of #207 and add a new UI element "Ignore Missing Dependencies" to control the strict existence checking of asset references before adding them to the job

This PR is built on top of #207 so the raw lines of code look higher until that PR is merged.
See [Ahuge - #29](https://github.com/Ahuge/deadline-cloud-for-after-effects/pull/29) for a PR of just the two branches in my repo
Additional option in the UI that will silence the existing popup when checked.

This UI option also drives a step parameter that sets `-continueOnMissingFootage` on the aerender launch command.

- [x] Testing using SMF from a Windows AE 25.3 client.
- [X] Testing using SMF from a Windows AE 24.6 client.
Yes

Not documented.

I don't believe so, by default the new feature is not enabled.

---

### DIFF
--- a/dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
+++ b/dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
@@ -111,6 +111,21 @@
       "default": 90
     },
     {
+      "name": "IgnoreMissingDependencies",
+      "type": "STRING",
+      "default": "OFF",
+      "allowedValues": [
+        "ON",
+        "OFF"
+      ],
+      "userInterface": {
+        "control": "DROPDOWN_LIST",
+        "label": "Ignore Missing Dependencies",
+        "groupLabel": "Ignore Missing Dependencies"
+      },
+      "description": "Missing dependencies rendering settings"
+    },
+    {
       "name": "JobScriptDir",
       "description": "Directory containing embedded scripts.",
       "userInterface": {
--- a/dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
+++ b/dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
@@ -123,7 +123,7 @@
         "label": "Ignore Missing Dependencies",
         "groupLabel": "Ignore Missing Dependencies"
       },
-      "description": "Missing dependencies rendering settings"
+      "description": "Allows render to continue without failing if referenced files are missing."
     },
     {
       "name": "JobScriptDir",
--- a/src/submission/JobTemplateHelper.jsx
+++ b/src/submission/JobTemplateHelper.jsx
@@ -78,6 +78,71 @@ function jobAttachmentsJson(inputFiles, outputFolder) {
     };
 }
 
+function processAVLayer(queue, exploredItems, attachments, ignoreMissingDependencies, shouldShowPopup) {
+    if (layer == null || !(layer instanceof AVLayer) || layer.source == null) {
+        return {
+            queue: queue,
+            exploredItems: exploredItems,
+            attachments: attachments,
+            shouldShowPopup: shouldShowPopup
+        }
+    }
+    var src = layer.source;
+    if (src.id in exploredItems) {
+        return {
+            queue: queue,
+            exploredItems: exploredItems,
+            attachments: attachments,
+            shouldShowPopup: shouldShowPopup
+        }
+    }
+    exploredItems[src.id] = true;
+    if (src instanceof CompItem) {
+        queue.push(src);
+    } else if (src instanceof FootageItem && src.mainSource instanceof FileSource) {
+        // We only care if the footage is missing when ignoreMissingDependencies is false
+        if (src.footageMissing && !ignoreMissingDependencies) {
+            if (shouldShowPopup) {
+                adcAlert(
+                    "Missing Footage: " +
+                    src.name +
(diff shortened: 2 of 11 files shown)


dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
```diff
@@ -110,6 +110,21 @@
...
+    {
+      "name": "IgnoreMissingDependencies",
+      "type": "STRING",
+      "default": "OFF",
+      "allowedValues": [
+        "ON",
+        "OFF"
+      ],
+      "userInterface": {
+        "control": "DROPDOWN_LIST",
```
eherozhao, line 122: https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#21-jobstringparameterdefinition

> “CHECK_BOX” — This is a checkbox control. It requires that allowedValues is also provided, and contains two values, case-insensitive, one representing true and another representing false.
Valid pairs are [“true”, “false”], [“yes”, “no”], [“on”, “off”], and [“1”, “0”].

check_box makes more sense to me here.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
```diff
@@ -110,6 +110,21 @@
...
+        "label": "Ignore Missing Dependencies",
+        "groupLabel": "Ignore Missing Dependencies"
+      },
+      "description": "Missing dependencies rendering settings"
```
eherozhao: can we make this description more specific about the rendering will not fail even if there are missing dependencies? Something like this.

src/submission/JobTemplateHelper.jsx
```diff
@@ -132,7 +143,8 @@ function findJobAttachments(rootComp) {
         // Notify the user if any fonts are missing or are substituted during the session.
         // A substituted font is a font that was already missing when the project is opened.
         // A missing font is a font that went missing (e.g. font was uninstalled) while the project was open.
-        if (app.fonts.missingOrSubstitutedFonts != "") {
+        //  Again only only care if ignoreMissingDependencies is false
```
eherozhao: duplicate only

olly-glenn-arch: Fixed

dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
olly-glenn-arch: Done

dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
olly-glenn-arch: It looks like the rest of the UI uses dropdowns, so it may be best to stick with a dropdown here to keep everything consistent. Also it looks like there's a bug with the `deadline bundle gui-submit` command where it won't read in the values of checkbox properties properly, which might explain why checkboxes weren't used in any other part of this section of the submitter UI (I followed up on Slack).

jaikish (dismissed):

dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
viknith: We have MFR as a checkbox right now, so why not keep this a checkbox as well?

olly-glenn-arch:
Testing clarification: I tested this PR by using an After Effects project that deliberately had an asset missing and kicking off two render jobs, one with ignore missing dependencies enabled and one with it disabled. When enabled, the comp rendered using the default media-offline After Effects asset in place of the missing one, and when disabled, the render job failed with an error saying that it couldn't find the file.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/parameter_definitions_image_fragment.json
olly-glenn-arch: This is referring to the external Deadline Cloud submission UI that pops up after you submit a job from inside After Effects. In the After Effects plugin, both ignore missing dependencies and multi-frame rendering are represented by checkboxes, and in the submitter popup both are represented by dropdowns. 
image

viknith: Got it

viknith (dismissed):

jaikish (approved):

viknith (approved):

Issue #30, opened 2024-05-03
Bug: Adaptor not working correctly on Windows CMF


yuanmich2:
### Expected Behaviour

I should be able to install python, install the aws cli, run pip install deadline deadline-cloud-for-after-effects deadline-cloud-worker-agent, and run the install-deadline-worker command to set up my worker.

### Current Behaviour

When I first set up my worker and ran a job, I was hit with this error:

```
2024/05/03 13:46:36-07:00 C:\ProgramData\Amazon\OpenJD\session-7703d39d3c31476e87a0691ecfbf6ba7y10ru5nk>afterfx-openjd daemon start --connection-file C:\ProgramData\Amazon\OpenJD\session-7703d39d3c31476e87a0691ecfbf6ba7y10ru5nk/connection.json --init-data file://C:\ProgramData\Amazon\OpenJD\session-7703d39d3c31476e87a0691ecfbf6ba7y10ru5nk\embedded_filesy89in_x0\init-data.yaml  
2024/05/03 13:46:36-07:00 Traceback (most recent call last):
2024/05/03 13:46:36-07:00   File "<frozen runpy>", line 198, in _run_module_as_main
2024/05/03 13:46:36-07:00   File "<frozen runpy>", line 88, in _run_code
2024/05/03 13:46:36-07:00   File "C:\Program Files\Python311\Scripts\afterfx-openjd.exe\__main__.py", line 4, in <module>
2024/05/03 13:46:36-07:00 ModuleNotFoundError: No module named 'deadline.ae_adaptor'
```

I found a folder called ae_adaptor at C:\Program Files\Python311\Lib\site-packages which seemed like the wrong directory, so I moved it to C:\Program Files\Python311\Lib\site-packages\deadline

then I got this error:
```
2024/05/03 13:49:45-07:00 INFO: Running command: python "C:\Program Files\Python311\Lib\site-packages\deadline\ae_adaptor\AEClient\ae_client.py"
2024/05/03 13:49:45-07:00  
2024/05/03 13:49:45-07:00 STDERR: Traceback (most recent call last):
2024/05/03 13:49:45-07:00  
2024/05/03 13:49:45-07:00 STDERR:   File "C:\Program Files\Python311\Lib\site-packages\deadline\ae_adaptor\AEClient\ae_client.py", line 111, in <module>
2024/05/03 13:49:45-07:00  
2024/05/03 13:49:45-07:00 STDERR:     main()
[...]

PR #214, opened 2025-06-27
fix: Only the first frame from an image sequence gets included in Job Attachments

### LINKED ISSUE #213: Bug: Only the first frame from an image sequence gets included in Job Attachments
### Describe the bug

If my composition contains footage that is an image sequence instead of a video, the submitter only adds the first frame from the image sequence to the job attachments input list.

Image

### Expected Behaviour

I would expect it to grab the entire frame range that is used

Image

### Current Behaviour

Only grabs the first frame

### Reproduction Steps

Import an Image sequence (png sequence for exmaple) and then submit that comp with JobAttachments

### Environment

At minimum:
1. Windows 11 Pro
2. AE 25.3 & AE 24.6
3. Latest

Ahuge:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/213

Only the first frame from an image sequence gets included in Job Attachments
When you have an image sequence in your comp, it should JA the entire used sequence otherwise you get missing footage issues.

Build upon the work of #207.

This updates the `findJobAttachments` function to use new logic built in `determineFootageType` and `filePathsFromFootageItem`

`determineFootageType` is a helper method used to determine the type of footage referenced. This was required because there is no native method in AE to determine if the FootageItem is an ImageSequence or Video.

`filePathsFromFootageItem` utilizes that function to implement logic against the ImageSequence footage items and gather all images in the folder from that sequence.
It takes inspiration from the `Smart Import.jsx` for resolving sequences with added logic around it to ensure that the files are part of the same sequence.
In addition to this, the function will gather the file paths starting from the first frame imported and only for the length of the FootageItem. The idea is that hopefully if your image sequence starts at frame 002 for example and is only used for 10 frames, we only upload the portion of the image sequence files on disk that are actually used in the comp.

This PR is built on top of #207 so the raw lines of code look higher until that PR is merged.
See [Ahuge - #30](https://github.com/Ahuge/deadline-cloud-for-after-effects/pull/30) for a PR of just the two branches in my repo
Additional input files in the Job Attachments screen if the composition uses Image Sequences.

- [X] Testing using SMF from a Windows AE 25.3 client.
- [x] Testing using SMF from a Windows AE 24.6 client.
Yes

Not documented as this fixes functionality that the user expected to exist

I don't believe so. This corrects functionality that was missing.

#### Before:
Job Attachments Before

#### After
Job Attachments After

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -77,7 +77,7 @@ var FootageTypes = {
     Audio: 3,
     Unknown: 4
 }
-var ImageExtensionRegex = new RegExp("(ai|bmp|bw|cin|cr2|crw|dcr|dng|dib|dpx|eps|erf|exr|gif|hdr|icb|iff|jpe|jpeg|jpg|mos|mrw|nef|orf|pbm|pef|pct|pcx|pdf|pic|pict|png|ps|psd|pxr|raf|raw|rgb|rgbe|rla|rle|rpf|sgi|srf|tdi|tga|tif|tiff|vda|vst|x3f|xyze)", "i")
+const ImageExtensionRegex = /"(ai|bmp|bw|cin|cr2|crw|dcr|dng|dib|dpx|eps|erf|exr|gif|hdr|icb|iff|jpe|jpeg|jpg|mos|mrw|nef|orf|pbm|pef|pct|pcx|pdf|pic|pict|png|ps|psd|pxr|raf|raw|rgb|rgbe|rla|rle|rpf|sgi|srf|tdi|tga|tif|tiff|vda|vst|x3f|xyze)/i;
 
 function readFile(filePath) {
     const f = new File(filePath);
@@ -340,7 +340,7 @@ function __generateUtil() {
          * @param {int} minValue - Minimum value that the slider/edittext can have.
          * @param {int} maxValue - Maximum value that the slider/edittext can have
          */
-        textObj.onChange = function() {
+        textObj.onChange = function () {
             const newValue = parseFloat(textObj.text);
             if (!isNaN(newValue) && newValue >= minValue && newValue <= maxValue) {
                 sliderObj.value = newValue;
@@ -349,7 +349,7 @@ function __generateUtil() {
         }
 
 
-        sliderObj.onChange = function() {
+        sliderObj.onChange = function () {
             textObj.text = Math.round(this.value);
             logger.log("Changed sliderObject(" + sliderObj.name + ") value to: " + Math.round(this.value), scriptFileUtilName, LOG_LEVEL.DEBUG);
         }
@@ -587,33 +587,28 @@ function __generateUtil() {
     }
 
     function filePathsFromFootageItem(footageItem) {
-        var paths = [];
+        const paths = [];
         if (determineFootageType(footageItem) === FootageTypes.ImageSequence) {
-            var source = footageItem.mainSource;
-            var frameCount = footageItem.duration / footageItem.frameDuration;
-            var firstFrame = new File(source.file.fsName).fsName;
+            const source = footageItem.mainSource;
+            const frameCount = footageItem.duration / footageItem.frameDuration;
+            const firstFrame = new File(source.file.fsName).fsName;
             logger.debug("Processing ImageSequence with (" + frameCount + ") frames: " + firstFrame);
-            var firstFramePattern = firstFrame.replace(new RegExp("[0-9]", "g"), "[0-9]").replace(new RegExp("\\\\", "g"), "\\\\");
-            var firstFrameRegex = new RegExp(firstFramePattern);
+            const firstFramePattern = firstFrame.replace(/\d/g, "\\d").replace(/\\\\/g, "\\\\");
+            const firstFrameRegex = new RegExp(firstFramePattern);
             logger.debug("  Regex pattern: " + firstFramePattern);
             var containingFolder = source.file.parent;
-
             function matchFilePattern(fileFolderObj) {
-                var fsName = fileFolderObj.fsName;
-                var result = fsName.match(firstFrameRegex);
-                return result
+                return fileFolderObj.fsName.match(firstFrameRegex);
             }
-            var containingFiles = containingFolder.getFiles(matchFilePattern).sort();
+            const containingFiles = containingFolder.getFiles(matchFilePattern).sort();
             logger.debug("  Pattern matched " + containingFiles.length + " files");
             var firstFrameFound = false;
             var indexOffset = 0;
             for (var index = 0; index < containingFiles.length; index++) {
                 var filePath = new File(containingFiles[index]).fsName;
-                if (firstFrameFound) {
-                    if ((index - indexOffset) < frameCount) {
-                        logger.debug("Adding frame (" + (index - indexOffset) + ") to paths: " + filePath);
-                        paths.push(filePath);
-                    }
+                if (firstFrameFound && ((index - indexOffset) < frameCount)) {
+                    logger.debug("Adding frame (" + (index - indexOffset) + ") to paths: " + filePath);
+                    paths.push(filePath);
                 }
                 if (filePath === firstFrame) {
                     firstFrameFound = true;
@@ -622,10 +617,8 @@ function __generateUtil() {
                     paths.push(filePath);
                 }
             }
-        } else {
-            if (footageItem.mainSource instanceof FileSource) {
-                paths.push(footageItem.mainSource.file.fsName);
-            }
+        } else if (footageItem.mainSource instanceof FileSource) {
+            paths.push(footageItem.mainSource.file.fsName);
         }
         return paths;
     }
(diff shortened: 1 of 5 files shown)


eherozhao (commented):
A way to make this CR not depending on the other ongoing PRs is to create a separate branch based on mainline for this feature dev. That can help reviewers focus on the real changes. If it is an easy switch, do you mind move the changes to a new branch and raise it up?

Ahuge:
Rebased off of mainline so you just see this PR

olly-glenn-arch:
For now I've rebased this off of #207 so that when #207 gets merged rebasing this branch off of mainline shouldn't be too messy. Just tested on Windows with Aftereffects 24.6.7, 25.1.0, and 25.2.2, and frame sequences rendered as expected. I've attached a more detailed breakdown of the tests I ran:

[_TESTING_NOTES.txt](https://github.com/user-attachments/files/21647335/_TESTING_NOTES.txt)

olly-glenn-arch:
Tested and working on AE 24.6.7, 25.1.0, and 25.2.2 on Windows, will test on Mac tomorrow.

src/utils/Utils.jsx
```diff
@@ -66,6 +66,15 @@ if (typeof DEFAULT_MAX_CPU_USAGE_PERCENTAGE === "undefined") {
...
 }
 
+var FootageTypes = {
+    Image: 0,
+    ImageSequence: 1,
+    Video: 2,
+    Audio: 3,
+    Unknown: 4
+}
+var ImageExtensionRegex = /(ai|bmp|bw|cin|cr2|crw|dcr|dng|dib|dpx|eps|erf|exr|gif|hdr|icb|iff|jpe|jpeg|jpg|mos|mrw|nef|orf|pbm|pef|pct|pcx|pdf|pic|pict|png|ps|psd|pxr|raf|raw|rgb|rgbe|rla|rle|rpf|sgi|srf|tdi|tga|tif|tiff|vda|vst|x3f|xyze)/i;
```
viknith: These are a lot of image extension types. Have you confirmed that all of them are supported for rendering on After Effects? I recall there are customers with their own custom image or video type that expect to render on After Effects. What would the user experience be here if they provide an image sequence that doesn't match this regex? Do we need to surface a warning or an info popup about it?

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -559,6 +568,65 @@ function __generateUtil() {
...
+        var regex = /^(.*?)(\d*)(\D*?)$/;
+        var match = fileName.match(regex);
+        return {
+            prefix: match[1],
+            frame: parseInt(match[2], 10),
+            suffix: match[3]
+        }
+    }
+
+    function filePathsFromFootageItem(footageItem) {
```
viknith: Nit: rename to getFilePathsFromFootageItem

src/utils/Utils.jsx
```diff
@@ -555,6 +564,65 @@ function __generateUtil() {
...
+     * Extracts image 
+     * @param {string} fileName 
+     * @returns Object containing prefix, name, and suffix 
+     */
+    function getImageSequenceInformation(fileName) {
+        var regex = /^(.*?)(\d*)(\D*?)$/;
+        var match = fileName.match(regex);
+        return {
+            prefix: match[1],
+            frame: parseInt(match[2], 10),
```
viknith, line 610: Does this represent the number of frames? If so, renaming to frameCount or numberOfFrames would probably be better, thoughts?

src/utils/Utils.jsx
olly-glenn-arch: Aftereffects doesn't differentiate between image sequences and video files when importing video footage, so this regex is necessary to figure out whether a video is an image sequence or not. Currently, the way the type of footage is determined is that we first query whether the source is a still image, and if it's not a still image then we check whether the file has one of these extensions to determine if it's an image sequence or video file (an image sequence is just non-still footage that has an image file extension). The ExtendScript API is very restrictive here so I'm not sure if there's any other way of doing it. The list of supported image file extensions is taken from here: https://helpx.adobe.com/after-effects/kb/supported-file-formats.html

If a user provides an image sequence that doesn't match this regex that means it's in an unsupported file format so the project theoretically should have broken before we got to this step. If a customer is using a proprietary image format / video codec then there would be no easy way for us to determine whether it was a still image sequence or video file (again because in ExtendScript there is no differentiation between the two) and supporting that case may fall outside the scope of this PR. It's also possible that the in-house image formats that customers are using are just extended versions of existing formats that share the same file extensions--do you have any examples of custom image file formats / video containers/codecs that customers are using? As long as the file extensions are in Adobe's list of supported extensions it should work fine.

dist/DeadlineCloudSubmitter.jsx
olly-glenn-arch: Done

src/utils/Utils.jsx
olly-glenn-arch: No, each of these objects represents an individual frame. Updated comment to clarify

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -559,6 +568,65 @@ function __generateUtil() {
...
+        return FootageTypes.Unknown
+    }
+
+    /**
+     * Extracts frame number, prefix, and suffix for single frame in image sequence.
+     * @param {string} fileName 
+     * @returns Object containing prefix, name, and suffix 
+     */
+    function getImageSequenceInformation(fileName) {
+        var regex = /^(.*?)(\d*)(\D*?)$/;
```
viknith: Security hotspot finding: https://sonarcloud.io/project/security_hotspots?id=aws-deadline_deadline-cloud-for-after-effects&pullRequest=214&hotspots=AZi-mbtK4c14mioZB6lB

olly-glenn-arch: It looks like the problem sonar is having here is that slow regex operations can create vulnerabilities for DDoS attacks. I updated the regex to make the last operator greedy instead of lazy which should help with performance, but the first operator has to be lazy because otherwise it would just match the entire string every time. Also, considering that this is running locally on the user's machine, I don't think a DDoS attack is something we'd need to worry about..

olly-glenn-arch:
Just tested on supported version of AE on both Windows and Mac and image sequence job attachments are being uploaded correctly, if there are no other changes to be made this should be ready to merge.

viknith (dismissed):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -70,6 +70,15 @@ if (typeof DEFAULT_MAX_CPU_USAGE_PERCENTAGE === "undefined") {
...
 }
 
+var FootageTypes = {
+    Image: 0,
+    ImageSequence: 1,
+    Video: 2,
+    Audio: 3,
+    Unknown: 4
+}
+var ImageExtensionRegex = /(ai|bmp|bw|cin|cr2|crw|dcr|dng|dib|dpx|eps|erf|exr|gif|hdr|icb|iff|jpe|jpeg|jpg|mos|mrw|nef|orf|pbm|pef|pct|pcx|pdf|pic|pict|png|ps|psd|pxr|raf|raw|rgb|rgbe|rla|rle|rpf|sgi|srf|tdi|tga|tif|tiff|vda|vst|x3f|xyze)/i;
```
eherozhao: https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/src/submission/JobTemplateHelper.jsx#L463 we have already had a similiar method why not just use it?

olly-glenn-arch: Good point, I'll switch to that method and also update it to include all of the image outputs that AE supports

olly-glenn-arch: Done

olly-glenn-arch:
Tested again and ready for review

viknith (approved):

viknith (dismissed):

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -559,6 +567,81 @@ function __generateUtil() {
...
+    // Return the `FootageTypes` value for the passed footageItem
+    function determineFootageType(footageItem) {
+        if (footageItem.hasVideo) {
+            var filePath = File.decode(footageItem.mainSource.file);
+            var extension = filePath.substr(filePath.lastIndexOf(".") + 1, filePath.length).toLowerCase();
+            if (footageItem.mainSource.isStill) {
+                return FootageTypes.Image
+            } else if (isImage(extension)) {
+                return FootageTypes.ImageSequence
+            } else {
```
jaikish, line 595: Should we check `isVideo` here?

olly-glenn-arch: This line of code is only reached if `footageItem.hasVideo` is true, so we already know that the footage item is a video. However, since AE considers both image sequences and still images to be special types of videos, we have to double-check the file extension to figure out if the video is actually a still or image sequence instead. If it isn't a still or image sequence, then it's a normal video.

jaikish (approved):

eherozhao (approved):

PR #223, opened 2025-07-11
fix: change the error handling for aerender cmd to fail the task when…

eherozhao:
We noticed that the call_aerender task didn't fail when the aerender error got thrown from the cmd.

1. Updated the error handling to fail the task.
2. Added more logging content.

This can be a pattern for other error patterns we would like to fail the task.

The task can be real failed when aerender throws the error.

I was able to submitted the job in AE 25 macOS with job thrown aerender error and job succeeded.
Examples:
image
image

Before this change, the failed task as image 1 looks as below:
image

Since this is the change for the python script running in the worker side, so we don't need to test it in different client side submitter.

N/A

N/A

N/A

N/A

---

### DIFF
--- a/dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
+++ b/dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
@@ -25,8 +25,18 @@ def main():
     )
     parser.add_argument("--chunk-size", type=int, help="The number of frames per task")
     parser.add_argument("--index", type=int, help="The starting frame of the chunk")
-    parser.add_argument("--multi-frame-rendering", type=str, default="OFF", help="Multi-frame render (MFR)")
-    parser.add_argument("--max-cpu-usage-percentage", type=int, default=90, help="Specifies the desired maximum CPU percentage power to use during rendering. Value is ignored if MFR is OFF")
+    parser.add_argument(
+        "--multi-frame-rendering",
+        type=str,
+        default="OFF",
+        help="Multi-frame render (MFR)",
+    )
+    parser.add_argument(
+        "--max-cpu-usage-percentage",
+        type=int,
+        default=90,
+        help="Specifies the desired maximum CPU percentage power to use during rendering. Value is ignored if MFR is OFF",
+    )
 
     args = parser.parse_args()
     print(f"Args: {args}", flush=True)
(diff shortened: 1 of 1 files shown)


jaikish (dismissed):

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -84,34 +94,80 @@ def main():
...
+        print(
+            f"[DEBUG] Process finished with return code: {process.returncode}",
+            flush=True,
+        )
 
-        # Also handle stderr (errors)
-        for line in process.stderr:
-            print(line, end="", flush=True)  # Print stderr line by line
+        # Process stdout line by line with debug markers
+        print("[DEBUG] Processing stdout...", flush=True)
```
npmacl: I don't think this print helps with much other than debugging when making this change, I think we should take it out.

eherozhao: deleted it.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/call_aerender.py
```diff
@@ -84,34 +94,78 @@ def main():
...
+                error_found = True
+
+        # If process returned non-zero code, that's also an error
+        if process.returncode != 0:
+            print(
+                f"[ERROR] Process returned non-zero exit code: {process.returncode}",
+                file=sys.stderr,
+                flush=True,
+            )
+            error_found = True
```
npmacl, line 151: not a huge fan of this approach, but also not sure if we have anything better since AE doesn't seem to be setting the return code to nonzero even when errors occur. My main concern would be erroring out on any false positives.

Seems like this person had a very similar approach back in 2021 to parse output for "ERROR": https://community.adobe.com/t5/after-effects-discussions/aerender-no-exit-code/m-p/12520419

Maybe to be cautious we should still parse for "ERROR" in the output of stderr? Im not sure if AE has any warning logs that might get written to stderr that we wouldn't want to fail on.

eherozhao: I am not too worried about it at this moment since we have above logic 
```
for line in stdout.splitlines():
            print(f"[STDOUT] {line}", flush=True)

            # Check for errors in stdout (case insensitive)
            line_lower = line.lower()
            if "error" in line_lower:
```
doing almost the same as that link was doing.

npmacl: what we are doing that the above link doesn't mention is failing the task if there is any output in stderr. my concern is if any warning level messages are sent to stderr we will fail a job that we shouldn't have, so I was suggesting we also have the ` if "error" in line_lower:` check when parsing stderr.

eherozhao: Updated.

npmacl (approved):

jaikish (approved):

PR #224, opened 2025-07-14
feat: add timeout checkbox for customers to set a timeout number for openjd

eherozhao:
We are having an issue that customer's task gets stuck without timed out.

Due to the limitation of openjd (rfc to improve https://github.com/OpenJobDescription/openjd-specifications/discussions/77), we have to create a checkbox for customers to provide the timeout parameter, which will be added to the task OnRun's timeout argument.

Screenshot 2025-07-14 at 2 35 04 PM

https://github.com/user-attachments/assets/58605b64-6985-4f89-bbc4-5cdadae28316

Customer's job will not be stuck forever.

I submitted the job from a macOS submitter and validated the generated video and image templates are having the timeout param with the right calculated seconds.

Yes

N/A

N/A

- Did you update all relevant docstrings for functions that you modified?
Yes
- Should the README.md, DEVELOPMENT.md, or other documents in the repository's docs/ directory be updated along with your change?

Yes, updated readme.md

N/A

---


src/submission/SubmitBundle.jsx
```diff
@@ -1,7 +1,16 @@
 /**
  * Submit the selected render queue item
  **/
-function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUsagePercentage) {
+function SubmitSelection(selection, framesPerTask, multiFrameRendering, maxCpuUsagePercentage, taskTimeoutDays, taskTimeoutHours, taskTimeoutMinutes) {
+    // Calculate task run timeout in seconds
+    var taskTimeoutSeconds = 0;
+    // Validate timeout values during job submission
+    if (taskTimeoutDays === 0 && taskTimeoutHours === 0 && taskTimeoutMinutes === 0) {
```
jaikish, line 8: should we do `<==0` to reflect what the error message is saying "must be greater than 0"?

src/ui/SubmitterUI.jsx
```diff
@@ -118,6 +118,85 @@ function buildUI(thisObj) {
...
+
+    // Function to validate timeout values
+    function validateTimeoutValues() {
+        // Check if all values are zero when checkbox is checked
+        if (taskRunCheckbox.value) {
+            var days = parseInt(taskRunDaysInput.text) || 0;
+            var hours = parseInt(taskRunHoursInput.text) || 0;
+            var minutes = parseInt(taskRunMinutesInput.text) || 0;
+
+            if (days === 0 && hours === 0 && minutes === 0) {
```
jaikish, line 160: Should we validate some upper bounds too?

jaikish (approved):
Couple of questions on if we could tighten the validations even further. Good to merge in current state to support the new feature!

src/submission/SubmitBundle.jsx
eherozhao: due to 
```
this.text = this.text.replace(/[^0-9]/g, "");
        if (this.text === "") this.text = "0";
```
we have above, the number will be always beyond or equal to 0.

src/ui/SubmitterUI.jsx
eherozhao: This is a good question. I thought about this when working on this. Since this will be the number passed to OpenJd timeout so I checked openJd timeout but didn't see a upper bound  https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md. That's why I didn't add a upper bound validation.

MarinMaksutaj (approved):

PR #227, opened 2025-07-16
ci: move installer build/release in release workflow

moorec-aws:
We want to build the installers and only release them after conda testing has been completed.

Use the [reusable_build_installers.yml](https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_build_installers.yml) prior to `IsCondaReady` to only build the installers. After the `IsCondaReady` step use [reusable_release_installers.yml](https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_release_installers.yml) to release the installers.

Re-orders the release process to better align with development needs.

https://github.com/moorec-aws/deadline-cloud-for-blender/actions/runs/16326741245

No

No

----

### DIFF
--- a/.github/workflows/release_publish.yml
+++ b/.github/workflows/release_publish.yml
@@ -27,20 +27,22 @@ jobs:
     with:
       tag: ${{ github.event_name == 'workflow_dispatch' && inputs.tag || '' }}
 
-  BuildAndStageInstallers:
-    needs: TagRelease
-    uses: aws-deadline/.github/.github/workflows/reusable_build_and_stage_installers.yml@mainline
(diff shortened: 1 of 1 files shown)


viknith (approved):

eherozhao:
Dumb question: this will not impact any release process operator need to operate right?

moorec-aws:
> Dumb question: this will not impact any release process operator need to operate right?

It does not change any processes from an operator perspective. It just makes sure the installer doesn't release to the build pipeline before testing occurs.

eherozhao (approved):

PR #228, opened 2025-07-18
fix: update start and end frame to reflect frame numbers accurately

viknith:
When customers would submit jobs that started with a non-zero frame number, the corresponding start frame number would still be 0. For example, if a composition was 100-399, it would show as 0-299 or something like that. This would cause problems when trying to inspect the output of specific frames.

Fix our logic for getting start and end frames so that the numbers are reflected accurately in the submitter and on the Deadline Cloud Monitor.

Resulting submitter view
Screenshot 2025-07-18 at 4 31 03 PM

The starting number used to be all zeroes but now it reflects the start frame of the composition accurately. When I submitted one of them, the tasks look like this:

Screenshot 2025-07-18 at 4 30 15 PM

Submitted jobs on Mac and Windows After Effects 25.2 and 24.6, verified it all works. 

Here were some cases I ran and submitted as PNG sequences:
- Job with 100-399 frames submitter with 24fps, 23.5fps, 23.75fps
- Job with 1-300 frame at 24fps
- Job with 0-299 frame at 24fps

Yes

No need

No

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -756,8 +756,7 @@ function __generateUtil() {
          */
         const startFrame = Number(rqi.comp.displayStartFrame);
         // Calculate number of frames using timeSpanDuration and frameRate
-        const timeSpanDuration = rqi.timeSpanDuration || rqi.comp.duration;
-        const numFrames = Math.floor(timeSpanDuration * rqi.comp.frameRate);
+        const numFrames = Math.floor(rqi.timeSpanDuration * rqi.comp.frameRate);
         const endFrame = startFrame + numFrames - 1; // end frame is inclusive
 
         return {
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -754,9 +754,9 @@ function __generateUtil() {
          * @param {RenderQueueItem} rqi - The render queue item to calculate frames for
          * @returns {Object} Object containing startFrame and endFrame
          */
-        const startFrame = Number(rqi.comp.displayStartFrame);
-        // Calculate number of frames using timeSpanDuration and frameRate
-        const numFrames = Math.floor(rqi.timeSpanDuration * rqi.comp.frameRate);
+         // NOTE: we're not using displayStartFrame since it is rounded up
+        const startFrame = Number(Math.floor(rqi.comp.displayStartTime * rqi.comp.frameRate));
+        const numFrames = Number(Math.floor(rqi.timeSpanDuration * rqi.comp.frameRate));
         const endFrame = startFrame + numFrames - 1; // end frame is inclusive
 
         return {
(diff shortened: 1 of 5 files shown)


src/utils/Utils.jsx
```diff
@@ -757,6 +744,24 @@ function __generateUtil() {
...
 
+    function calculateFrameRange(rqi) {
+        /**
+         * Calculate start and end frames for a render queue item using render settings
+         * @param {RenderQueueItem} rqi - The render queue item to calculate frames for
+         * @returns {Object} Object containing startFrame and endFrame
+         */
+        const startFrame = Number(rqi.comp.displayStartFrame);
+        // Calculate number of frames using timeSpanDuration and frameRate
+        const timeSpanDuration = rqi.timeSpanDuration || rqi.comp.duration;
```
eherozhao: q: what's the diff between these two?

viknith: Good call, I missed this. Took a closer look at documentation and the CompItem doesn't have a duration field. Will revise this.

dist/DeadlineCloudSubmitter.jsx
```diff
@@ -761,6 +748,23 @@ function __generateUtil() {
...
 
+    function calculateFrameRange(rqi) {
+        /**
+         * Calculate start and end frames for a render queue item using render settings
+         * @param {RenderQueueItem} rqi - The render queue item to calculate frames for
+         * @returns {Object} Object containing startFrame and endFrame
+         */
+        const startFrame = Number(rqi.comp.displayStartFrame);
+        // Calculate number of frames using timeSpanDuration and frameRate
+        const numFrames = Math.floor(rqi.timeSpanDuration * rqi.comp.frameRate);
```
eherozhao: https://ae-scripting.docsforadobe.dev/item/compitem/?h=displaystartframe#compitemframeduration Will this just be the frames of the comp?

viknith: This represents the number of frames in the comp, yes, rather than the number of seconds.

eherozhao (dismissed):

dist/DeadlineCloudSubmitter.jsx
npmacl: have we tested that this logic works correctly for fractional framerates, e.g. 29.97 FPS?

npmacl: Have we tested that this logic is correct for fractional frame rates, e.g. 29.97 FPS?

viknith: Good call, looking into it

viknith:
I used a decimal frames per task and noticed the start frame was off by a frame, so I modified the code to calculate the start frame as well.

dist/DeadlineCloudSubmitter.jsx
eherozhao: Can we add some test cases you have run in the PR description?

eherozhao (approved):

jaikish (approved):

PR #252, opened 2025-09-25
fix: moved font search logic to get_user_fonts.py to permanently prevent JSON size overflow issues that crashes After Effects

viknith:
On the rare case that the font used in an Adobe After Effects composition does not have its file location defined in the composition, the `get_user_fonts.py` script serves as a backup that gets all of the font files in the artist's machine and returns a JSON object to the After Effects submitter, which then iterates through the returned JSON object to get the target font's file path.

The problem is, the JSON object returned from the python file is massive and if an artist were to have a extremely high number of fonts on their machine, then After Effects will take forever to accept the JSON object and sometimes even crashes. 

To prevent this issue of returning a super large JSON object from the Python file that crashes After Effects, I moved the font file searching logic to the Python file itself and that file will just return the font file path that was found. 

Prevents AE crashes

Tested on AE 24/25 on Mac and AE 24/25 on Windows with the fallback script forced to run by removing the usage of `font.location`, and verified it worked on all OS and versions and jobs succeeded.

Yes

N/A

N/A

No

No

If so, then please describe the changes that users of this package must make to update their scripts, or Python applications. Also,
please ensure that the title of your commit follows our conventional commit guidelines in
[CONTRIBUTING.md](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/CONTRIBUTING.md#conventional-commits) for breaking changes.
_delete text ending here_

---

### DIFF
--- a/dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
+++ b/dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
@@ -1,38 +1,16 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-"""
-Collects font name metadata from user-installed fonts. Output as JSON.
-"""
-
-import json
 import os
 import sys
-import traceback
 
 try:
     from fontTools import ttLib
 except ModuleNotFoundError:
-    # Attempt to install fonttools automatically
-    try:
-        print("fonttools module not found. Attempting to install...")
-        import subprocess
-        subprocess.check_call([sys.executable, "-m", "pip", "install", "fonttools"])
-        print("fonttools installed successfully. Importing...")
-        from fontTools import ttLib
-    except Exception:
-        print(json.dumps({
-            "message": "Unable to install fonttools module. Please install manually with 'pip install fonttools' and try again.",
-            "error": traceback.format_exc()
-        }))
-        sys.exit(1)
-except Exception as e:
-    print(json.dumps({
-        "error": traceback.format_exc()
-    }))
-    sys.exit(1)
+    import subprocess
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "fonttools"])
+    from fontTools import ttLib
 
-
-# Font locations to search on Windows
+# Font locations to search
 SEARCH_PATHS = [
     "%APPDATA%/Adobe/CoreSync/plugins/livetype",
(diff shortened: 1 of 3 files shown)


dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -1,179 +1,64 @@
...
-        if not os.path.exists(root_path):
-            return result
-    except Exception:
-        if verbose:
-            print(traceback.format_exc())
-        return result
-
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        sys.exit(1)
```
npmacl, line 36: should we provide error message first?

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
```diff
@@ -1,179 +1,64 @@
...
-    their metadata.
-    Returns a dictionary with file paths as the keys.
-    """
-
-    fonts = {}
-
-    for search_path in search_paths:
-        search_root = os.path.normpath(os.path.expandvars(os.path.expanduser(search_path)))
-        try:
+        for search_path in SEARCH_PATHS:
```
npmacl, line 41: how long does this take? does it run every submission? dont think it needs to be part of the PR but doing this in parallel might make the submitter feel more responsive.

viknith: It only runs on the rare edge case where `font.location` is not defined, so it almost never adds latency, and its usually very quick. A future optimization can be to run this Python script once and then getting all of the locations in one run, but that can be done later.

dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py
viknith: Added!

npmacl (approved):

eherozhao:
> Tested on AE 24/25 on Mac and AE 24/25 on Windows with the fallback script forced to run by removing the usage of font.location, and verified it worked on all OS and versions and jobs succeeded.

We should investigate auto testing for this.

eherozhao (approved):

PR #256, opened 2025-10-15
fix: aggregate font errors into single popup for submission

jaikish:
Fixes:  Tracked internally

Current AE submitter shows one popup per font issue detected on user's workstation. This hinders their submission workflow when the number of unsupported/location-not-found fonts are high in number. We need to show a single popup.

Aggregated all errors into one popup per issue type for non-legacy AE versions. 

1. Created a separate method for handling error messages in non-legacy versions.
2. Retained multiple popup experience in legacy versions of AE(<24.5) to bias for action and resolve immediate user issues.

Users just need to acknowledge one popup before submission per issue type.

1. Reproduced the issue on AE 25.5 by creating a composition with 3 ttc type fonts and submitting. (3 popups shown).
3. Fixed the submitter and retried the submission. Verified that all errors are aggregated into single popup.

SinglepopupForFontIssues

Yes

NA

NA
NA

No

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1635,6 +1635,8 @@ function findJobAttachments(rootComp, ignoreMissingDependencies) {
  **/
 function getFontsFromFile() {
     var fontLocations = [];
+    var unsupportedFonts = [];
+    var fontsWithoutLocation = [];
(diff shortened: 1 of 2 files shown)


viknith:
Helvetica-Light and Helvetica-Oblique are actually are fonts that reside within Helvetica.ttc

PR #257, opened 2025-10-16
fix: aggregate font errors into single popup for submission

jaikish:
Fixes:  Tracked internally

Current AE submitter shows one popup per font issue detected on user's workstation. This hinders their submission workflow when the number of unsupported/location-not-found fonts are high in number. We need to show a single popup.

Aggregated all errors into one popup per issue type for non-legacy AE versions. 

1. Created a separate method for handling error messages in non-legacy versions.
2. Retained multiple popup experience in legacy versions of AE(<24.5) to bias for action and resolve immediate user issues.

Took @viknith 's advice to aggregate by font family in new PR

Users just need to acknowledge one popup before submission per issue type.

1. Reproduced the issue on AE 25.5 by creating a composition with 3 ttc type fonts and submitting. (3 popups shown).
4. Fixed the submitter and retried the submission. Verified that all errors are aggregated into single popup.

multipopup issue

Yes

NA

NA
NA

No

---

### DIFF
--- a/src/submission/JobTemplateHelper.jsx
+++ b/src/submission/JobTemplateHelper.jsx
@@ -201,6 +201,9 @@ function findJobAttachments(rootComp, ignoreMissingDependencies) {
  **/
 function getFontsFromFile() {
     var fontLocations = [];
+    var unsupportedFonts = {}; // using this object as a set because AE doesn't support sets
+    var unsupportedFontList = [];
+    var fontsWithoutLocation = [];
     // app.project.usedFonts was introduced in 24.5. Fall back to scanning text layers if version is older
     if (dcUtil.getAEVersion() >= 24.5) {
         const usedList = app.project.usedFonts;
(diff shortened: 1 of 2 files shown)


src/submission/JobTemplateHelper.jsx
```diff
@@ -362,6 +383,49 @@ function createFontFilename(fontLocation, fontPostScriptName) {
     return fontName;
 }
 
+
+/**
+ * Generates a font filename based on the font name and the extension of the font filename
+ * and whether the extension is supported
+ * @return an object with the font filename and extension validity
+ **/
+function getFontFilenameAndSupportStatus(fontLocation, fontPostScriptName) {
```
MarinMaksutaj, line 392: Is this the same function as above? Can we refactor somehow?

MarinMaksutaj (approved):

src/submission/JobTemplateHelper.jsx
jaikish: Hey, the above function is used in getFontsFromFileLegacy dedicated to legacy AE versions.

That was my reference in the Solution section of the description.
> Retained multiple popup experience in legacy versions of AE(<24.5) to bias for action and resolve immediate user issues.

eherozhao (approved):
not a blocker but if the list of unsupported font gets very large, will the UI become very long?

PR #259, opened 2025-10-16
fix: bug fix for valid fonts not being collected

jaikish:
Fixes: Tracked internally

In an [earlier PR](https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/257/files) to fix multiple popups per font issue, we introduced a regression which was caught by @viknith during testing for one of the positive scenarios. This was due to field refactoring in a function return value in the revised PR.

This was causing valid fonts not be collected by submitter

Fixed the return value so that the valid fonts are collected by submitter

Bug fix. Submitter collects all valid fonts for job attachments.

Created a comp with valid and invalid fonts and it showed popup for invalid fonts and collected valid fonts as job attachments.

Screenshot 2025-10-16 at 11 29 18 AM

Yes

NA

No

---

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1848,14 +1848,14 @@ function getFontFilenameAndSupportStatus(fontLocation, fontPostScriptName) {
     if (validExtension) {
         return {
             "isExtensionSupported": true,
-            "fontPostScriptName": fontPostScriptName,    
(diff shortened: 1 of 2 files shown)


eherozhao (approved):

viknith:
Validated that this works on Windows AE 24.6 and 25.4
Screenshot 2025-10-16 at 12 05 15 PM

viknith (approved):

PR #260, opened 2025-10-16
fix: improve get_user_fonts.py script and refactored file existence validation to minimize popups

viknith:
There were a series of minor issues related to fonts:
- There was another location on macOS machines with font files in it that were not being searched in the `get_user_fonts.py` fallback search script, called `/System/Library/Fonts`
- If a customer was using a `.ttc` font in their composition and the fallback script was run, the fallback script wouldn't be able to identify the font file used in the composition and thus wouldn't be able to warn the artist about using an incompatible font.
- If a customer had `get_user_fonts.py` missing, they ran the risk of getting the popup that it was missing several times.
- We had no unit testing of our `get_user_fonts.py` fallback script.

- Added the additional font folder location and improved fallback script to handle searching for .ttc files, and refactored asset file existence checks to happen on submit to allow the submitter to fail early.
- The diff is a bit larger because I ran `hatch run lint` to format all the Python files. 

Better fallback mechanism.

On macOS AE24.6 and AE 25.5, I attempted to submit a job on macOS and forced the fallback script to be run by doing a manual code change. I verified the script worked and was able to find the font Helvetica.ttc which it couldn't find before, and a warning correctly surfaced to me that Helvetica.ttc was not a supported file type. There's another font called TamilSangamMN that wasn't being identified before but now it's being identified properly.

Then I wrote unit tests which passed on macOS. However, I'm unable to run the unit tests on Windows at this time due to technical issues, still troubleshooting.

No

No

No

---

### DIFF
--- a/src/submission/SubmitBundle.jsx
+++ b/src/submission/SubmitBundle.jsx
@@ -238,12 +238,17 @@ function SubmitSelection(selection, selectionSettings) {
         "step_video_fragment.json"
     ];
 
+    var missingFiles = [];
     for (var f = 0; f < requiredFiles.length; f++) {
-        var jobTemplateFile = new File(assetsFolder.fsName + "/" + requiredFiles[f]);
-        if (!jobTemplateFile.exists) {
-            adcAlert("Error: Missing required file: " + requiredFiles[f], true);
-            return;
+        var requiredFile = new File(assetsFolder.fsName + "/" + requiredFiles[f]);
+        if (!requiredFile.exists) {
+            missingFiles.push(requiredFiles[f]);
         }
+    }
+    
+    if (missingFiles.length > 0) {
+        adcAlert("Error: Missing required files:\n" + missingFiles.join("\n"), true);
+        return;
     }
 
     const renderQueueItems = [];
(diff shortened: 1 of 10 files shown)


src/submission/SubmitBundle.jsx
```diff
@@ -216,6 +216,36 @@ function SubmitSelection(selection, selectionSettings) {
...
+        "video_template.json",
+        "job_environments_fragment.json",
+        "parameter_definitions_image_fragment.json",
+        "parameter_definitions_video_fragment.json",
+        "step_image_fragment.json",
+        "step_video_fragment.json"
+    ];
+
+    for (var f = 0; f < requiredFiles.length; f++) {
+        var jobTemplateFile = new File(assetsFolder.fsName + "/" + requiredFiles[f]);
```
jaikish: nit: not every required file is a `jobTemplateFile`. May be just `requiredFile`? and one single alert?

jaikish (dismissed):

src/submission/SubmitBundle.jsx
viknith: Good call, I can revise this.

jaikish (approved):

MarinMaksutaj (approved):

Issue #282, opened 2026-01-14
Bug: Frame range calculation does not account for project's displayStartFrame setting


MisakiAkatsuki:
### Describe the bug

The [calculateFrameRange()](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/2340a5fa36c952e1528488fcc68aa589753c169e/dist/DeadlineCloudSubmitter.jsx#L1011) function in `DeadlineCloudSubmitter.jsx` (line 1011-1025) does not account for the project's `app.project.displayStartFrame` setting when calculating frame ranges for render queue items.

When users configure their After Effects project to start at a non-zero frame number (e.g., frame 1 or frame 1001), the submitted job uses incorrect frame numbers that don't match what's displayed in the After Effects UI. This results in frame range mismatches and potentially incorrect or incomplete renders.

I have tested this with a project configured to start at frame 1 (`displayStartFrame = 1`) and confirmed that the frame range output is incorrect. The submitted job uses frame numbers starting from 0 instead of 1.

### Expected Behaviour

The frame range calculation should incorporate `app.project.displayStartFrame` to accurately reflect the project's frame numbering configuration. When a project is set to start at frame 1, the submitted job should also use frame numbers starting from 1, matching exactly what the user sees in the After Effects UI.

The calculation should be:

```javascript
const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate)) + app.project.displayStartFrame;
```

### Current Behaviour

The current implementation assumes all projects start at frame 0:

```javascript
const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate));
```

This causes the following issues:

- Jobs submitted from projects with `displayStartFrame = 1` use frame numbers starting from 0 instead of 1
- The frame numbers in the submitted job don't match the frame numbers shown in the After Effects UI
- This mismatch can cause incorrect frames to be rendered, potentially resulting in missing or wrong frames in the final output

For example, if a user sets up a render queue item to render frames 1-100 in the After Effects UI, the submitted job will actually render frames 0-99.

**Note:** The existing comment in the code states "we're not using displayStartFrame since it is rounded up", but this refers to `rqi.comp.displayStartFrame` (composition-level), not `app.project.displayStartFrame` (project-level). The project-level setting is a simple integer offset and should be respected.

### Reproduction Steps

[...]

PR #283, opened 2026-01-15
fix: add project displayStartFrame to frame range calculation

MisakiAkatsuki:
Fixes: #[[282](https://github.com/aws-deadline/deadline-cloud-for-after-effects/issues/282)]

When After Effects projects are configured with a non-zero `displayStartFrame` setting (e.g., starting at frame 1 or frame 1001), the Deadline Cloud Submitter calculates incorrect frame ranges for job submissions.

**Issue:**
- The `calculateFrameRange()` function does not account for `app.project.displayStartFrame`
- Jobs submitted from projects with custom start frames use frame numbers that don't match the After Effects UI display
- Example: A project with `displayStartFrame = 1` showing frames 1-240 in the UI gets submitted as frames 0-239

This causes frame range mismatches and potential incorrect or incomplete renders.

Modified the `calculateFrameRange()` function in `DeadlineCloudSubmitter.jsx` to add the `app.project.displayStartFrame` offset:

```javascript
const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate)) + app.project.displayStartFrame;
```

The `endFrame` automatically adjusts due to its dependency on `startFrame`, so no additional changes are needed.

 **Backward Compatible**: Projects with default `displayStartFrame = 0` see no change in behavior

 **Fixes Bug**: Projects with custom start frames now submit correct frame ranges that match the After Effects UI

 **No Breaking Changes**: Existing workflows continue to work as expected

- Tested with `displayStartFrame = 0` (default): Frame ranges unchanged
- Tested with `displayStartFrame = 1`: Frame range now correctly starts at 1 instead of 0
- Tested with `displayStartFrame = 1001`: Frame range now correctly starts at 1001 instead of 0
- Verified total frame count remains consistent across all configurations

Yes - Ran the bundler to regenerate `dist/DeadlineCloudSubmitter.jsx`

#### Was this change documented?

Updated the NOTE comment in the code to clarify the distinction between `rqi.comp.displayStartFrame` (composition-level) and `app.project.displayStartFrame` (project-level).

No - This is a backward-compatible bug fix that only affects projects using non-default `displayStartFrame` values, correcting their behavior to match user expectations.

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -1014,8 +1014,7 @@ function __generateUtil() {
          * @param {RenderQueueItem} rqi - The render queue item to calculate frames for
          * @returns {Object} Object containing startFrame and endFrame
          */
-        // NOTE: we're not using displayStartFrame since it is rounded up
-        const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate));
+        const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate)) + app.project.displayStartFrame;
         const numFrames = Number(Math.ceil(rqi.timeSpanDuration * rqi.comp.frameRate));
         const endFrame = startFrame + numFrames - 1; // end frame is inclusive
 
--- a/src/utils/Utils.jsx
+++ b/src/utils/Utils.jsx
@@ -1010,8 +1010,7 @@ function __generateUtil() {
          * @param {RenderQueueItem} rqi - The render queue item to calculate frames for
          * @returns {Object} Object containing startFrame and endFrame
          */
-        // NOTE: we're not using displayStartFrame since it is rounded up
-        const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate));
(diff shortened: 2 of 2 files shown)


dist/DeadlineCloudSubmitter.jsx
```diff
@@ -1014,8 +1014,7 @@ function __generateUtil() {
          * @param {RenderQueueItem} rqi - The render queue item to calculate frames for
          * @returns {Object} Object containing startFrame and endFrame
          */
-        // NOTE: we're not using displayStartFrame since it is rounded up
-        const startFrame = Number(Math.floor((rqi.comp.displayStartTime + rqi.timeSpanStart) * rqi.comp.frameRate));
```
eherozhao, line 1018: https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/2340a5fa36c952e1528488fcc68aa589753c169e/src/utils/Utils.jsx#L1014 is the original place we had this code. This file in dist/ is generated from the python script. If we now run the python script, it will be reverted back to the older version since utils.jsx is not updated. Could you update utils file and run the script to generate this jsx?

MisakiAkatsuki:
I've updated Utils.jsx. Is this correct?

eherozhao:
The PR looks good now. 
According to the screenshot of my AE 2025, it has start at 0 and start at 1 options, but in the PR description, you mentioned "or frame 1001", may I ask how will you be able to set the start frame more than 1?
image

eherozhao (approved):

eherozhao:
> I've updated Utils.jsx. Is this correct?


The PR looks good now. 👍

xuetinp (approved):

PR #285, opened 2026-01-27
docs: add user guide doc

eherozhao:
AE doesn't have a user guide yet
Created one by README.md content
Customer will be able to use user guide (the central page) to understand how to setup AE submitter
follow https://github.com/aws-deadline/aws-deadline.github.io to deploy it in local server Screenshot 2026-01-28 at 11 39 34 AM unfortunately it cannot be shared.
N/A

N/A
N/A
Y

N/A
---

### DIFF
--- a/docs/user_guide/installation.md
+++ b/docs/user_guide/installation.md
@@ -1,0 +1,87 @@
+# Installation
+
+To install AWS Deadline Cloud submitter for After Effects, please prepare the following environment:
+
+- Windows or macOS workstation
+- Adobe After Effects 24 or 25 installation
+- Python 3.9 or higher
+- Access to an AWS Deadline Cloud farm with either
+    - a service-managed fleet with After Effects available (via custom Conda package)
+    - a customer-managed fleet with After Effects and licensing set up
+
+## Prerequisites
+
+Before installing the submitter, you need to:
+
+1. Install the Deadline CLI and Deadline Cloud Monitor by running the Deadline Submitter and Deadline Monitor installers from the downloads section of the Deadline Cloud service in your AWS Console.
+
+2. Enable script permissions in After Effects. This submitter requires the ability to write files and send communication over the network in order to function properly. By default, After Effects scripts are not allowed to perform these actions. [Reference link](https://helpx.adobe.com/after-effects/using/scripts.html).
+
+   To allow scripts to write files or send communication over a network:
+   
+   - **Windows**: Select `Edit > Preferences > Scripting & Expressions > select Allow Scripts To Write Files And Access Network`
+   - **macOS**: Select `After Effects > Settings > Scripting & Expressions > select Allow Scripts To Write Files And Access Network`
+
+## Installing the submitter
+
+The After Effects submitter extension allows you to submit jobs to Deadline Cloud directly from within After Effects.
+
+**Note: During the installer process, you will choose between User Install or System Install. macOS users will default to User Install, since automatic System Install is currently not supported on macOS (manual System Install is still possible for admin users). Read only the section below that matches your selection:**
+
+### User Install
+
+1. Download the Deadline Cloud Submitter installer by following [Step 1: Install the Deadline Cloud Submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html#submitter-installation).
+2. Run the installer (no admin required).
+3. Follow the prompts and select the After Effects submitter and choose the major version (e.g., 24 or 25).
+4. The submitter will be installed based on your operating system:
+   - **macOS**: The installer automatically places `DeadlineCloudSubmitter(User).jsx` and `DeadlineCloudSubmitter_Assets` into your After Effects user preferences directory for all minor versions of the selected major version:
+     ```
+     /Users/<user>/Library/Preferences/Adobe/After Effects/<version>/Scripts/ScriptUI Panels
+     ```
+     The submitter will appear in the **Window** menu in After Effects.
+   - **Windows**: The submitter is installed to a standalone location by default:
+     ```
+     C:\Users\<user>\DeadlineCloudSubmitter\Submitters\AfterEffects\AE<version>
+     ```
+     You will need to run the script manually via **File > Scripts > Run Script File**.
+
+### System Install
+
+**Note: Automatic macOS system install is not yet supported by the submitter installer. macOS admin users can perform a manual system install by copying files to the application directory (see [Manual Installation](#manual-installation-alternative)).**
+
+1. Download the Deadline Cloud Submitter installer by following [Step 1: Install the Deadline Cloud Submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html#submitter-installation).
+2. **Windows only**: Right-click the installer and choose `Run as Admin`.
+3. Follow the prompts and select the After Effects submitter. It will be installed to:
+   - **Windows**: `C:\Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`
+   - **macOS** (manual): `/Applications/Adobe After Effects <version>/Scripts/ScriptUI Panels`
+
+### Final Setup Steps
+
+1. After installing via submitter installer, ensure you log into your Deadline user profile by running `deadline auth login` in the Terminal/Powershell or logging in via the Deadline Cloud Monitor.
+2. Next, to install the necessary dependencies used by the AE submitter, run the following in your local Terminal or Command Prompt:
+   ```
+   pip install fonttools
+   ```
+3. Finally, restart After Effects if it was open.
+
+## Manual Installation (Alternative)
+
+If you prefer to install manually, you can copy the submitter files directly:
+
+1. Locate the `DeadlineCloudSubmitter.jsx` file and the `DeadlineCloudSubmitter_Assets` folder in the `dist` folder of this repository.
+2. Copy both to the **ScriptUI Panels** folder within your After Effects installation:
+   - **Windows**: `Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`
+   - **macOS**: `Applications/Adobe After Effects <version>/Scripts/Script UI Panels`
+3. Restart After Effects if it was open.
+
+## Setting up After Effects with your Deadline Cloud Farm
+
+After Effects conda packages are available in AWS Deadline Cloud Service Managed Fleet. For the list of supported versions, see the [Deadline Cloud User Guide](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/create-queue-environment.html). If you would like to build a conda channel that contains a different After Effects conda package, please follow [these instructions](https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/configure-jobs-s3-channel.html).
+
+You can also use the After Effects conda recipe in [deadline-cloud-samples package](https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/conda_recipes/aftereffects-25.0) as a reference when building the package.
+
+Jobs created by this submitter require the `aerender` executable to be available on the PATH of the user that will be running your jobs. Alternatively, you can set the `AERENDER_EXECUTABLE` environment variable to point to the aerender executable.
+
+## Updating the submitter
+
+To update the submitter to the latest version, download and run the latest submitter installer.
--- a/docs/user_guide/using-submitter.md
+++ b/docs/user_guide/using-submitter.md
@@ -1,0 +1,178 @@
+# Using the After Effects Submitter
+
+To use the Deadline Cloud submitter for After Effects, please ensure your Farm is configured with an After Effects capable fleet (see [Setting up After Effects with your Deadline Cloud Farm](installation.md#setting-up-after-effects-with-your-deadline-cloud-farm)), and have the submitter installed. Also, please log into the Deadline Cloud Monitor or provide AWS credentials via a configuration profile for Deadline Cloud access.
+
+Refer to the [installation.md](installation.md) to configure the fleet and After Effects application.
+
+## Submit a job
+
+**Note: Follow the instructions below that match the install type you selected during installation.**
+
+### User Install - macOS
+
+1. Launch Adobe After Effects.
+2. Open the submitter by clicking **Window > DeadlineCloudSubmitter(User).jsx**.
+3. Follow the steps in [Submitting Your Job](#submitting-your-job) below.
+
+### User Install - Windows
+
+1. Launch Adobe After Effects.
+2. Open the submitter by clicking **File > Scripts > Run Script File** and navigate to where the `DeadlineCloudSubmitter.jsx` file is located and select it to run the submitter. If you recently closed the submitter, you can reopen it by clicking **File > Scripts > Recent Script Files** and pick the `DeadlineCloudSubmitter.jsx` file that was previously run.
+3. Follow the steps in [Submitting Your Job](#submitting-your-job) below.
+
+### System Install
+
+1. Launch After Effects as Admin (Windows) or with appropriate permissions (macOS).
+2. Open the submitter by clicking **Window > DeadlineCloudSubmitter.jsx**.
+3. Follow the steps in [Submitting Your Job](#submitting-your-job) below.
+
+### Submitting Your Job
+
+1. Click "Open Render Queue" on the submitter. Add any composition to your render queue and set up your render settings, output module, and output path.
+2. Click **Refresh** on the submitter to see your composition in the composition list.
+3. Select your composition you want to render and click **Submit** to submit a render job. For details on available settings, see [After Effects Specific Settings](#after-effects-specific-settings) below.
+4. If you see a warning popup window with "You are about to run the script contained in file", you can suppress the warning by following the instruction in the popup or the instructions in [installation.md](installation.md) to disable warnings when submitting jobs.
+5. Install any python libraries if prompted and press the **Login** button in the bottom left if you are not logged in.
+6. Set the farm and queue you are submitting to with the **Settings** button, and click **Submit**.
+
+**Note**: The After Effects submitter calls the Deadline GUI Submitter to complete job submission. If you hit any issues on the GUI submitter, please refer to [deadline-cloud](https://github.com/aws-deadline/deadline-cloud) library for help.
+
+### Shared Job Settings
+
+Settings that apply to the entire job:
+
+- **Farm Selection** - Choose which farm your job will render on
+- **Queue Selection** - Select the specific queue within your chosen farm
+- **Job Name** - Give your render job a descriptive name
+- **Job Description** - Add optional details about your render job
+- **Priority** - Set job priority for queue management
+- **Initial State** - Control whether the job starts immediately or remains paused
+- **Max Failed Tasks Count** - Maximum number of tasks that can fail before the job is marked as failed
+- **Max Retries Per Task** - Number of times a failed task will be retried
+- **Max Worker Count** - Maximum number of workers that can work on this job simultaneously
+
+### After Effects Specific Settings
+
+Settings specific to After Effects rendering:
+
+- **Composition Selection** - Select which composition from your render queue to submit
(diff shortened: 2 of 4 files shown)


docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+Refer to the [installation.md](installation.md) to configure the fleet and After Effects application.
+
+## Submit a job
+
+**Note: Follow the instructions below that match the install type you selected during installation:**
+
+### User Install
+
+1. Launch Adobe After Effects.
+2. Open submitter by clicking **File > Scripts > Run Script File** and navigate to where the `DeadlineCloudSubmitter.jsx` file is located and select it to run the submitter. If the submitter is closed, reopen it easily by clicking **File > Scripts > Recent Script Files** and pick the `DeadlineCloudSubmitter.jsx` file that was previously run.
```
npmacl: "Open submitter" -> "Open the submitter".

also I would reword "If the submitter is closed, reopen it easily" to "If you recently closed the submitter, you can reopen it by"

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+3. Click "Open Render Queue" on the submitter. Add any composition to your render queue and set up your render settings, output module, and output path.
+4. Then click **Refresh** on the submitter to see your composition in the composition list.
+5. Select your composition you want to render and click **Submit** to submit a render job. Here are some settings you can set:
+   - (Optional) For image sequences output types you can specify the number of frames per task so that the job created by the After Effects submitter will create the tasks based on the chunk size of the frames and then Deadline Cloud will assign the tasks to available workers to delegate the load.
```
npmacl: this sentence is pretty hard to follow. I would reword to something like

"For image sequences output types you can specify the number of frames per task. By default, Deadline Cloud creates one task for every frame. This setting changes this behavior to instead render the specified number of frames per task. This can significantly speed up rendering in jobs where each frame can be rendered relatively quickly". 

although if you can, I'd make the last sentence less vague than I did with "significantly speed up" and "relatively quickly", by adding some numbers if you can

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+   - You can also specify multi-frame rendering with your job submission. If you do, you can also specify the max percentage of CPU usage you wish to allocate towards rendering in case you would like to limit it to allow other background applications or processes to run smoothly. For more information about multi-frame rendering, visit Adobe's website [here](https://helpx.adobe.com/after-effects/using/multi-frame-rendering.html).
```
npmacl: >  If you do, you can also specify the max percentage of CPU usage you wish to allocate towards rendering in case you would like to limit it to allow other background applications or processes to run smoothly

do we recommend this at all? I think especially on SMF customers would get the most out of targeting 100%, so we should probably mention that.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+   - You need to set up timeout days, hours and minutes to avoid the task getting stuck forever. The default timeout is 2 days.
```
npmacl: is this needed? the timeouts are optional for nuke at least iirc. 

I would also reword a little bit, right now it reads like "your task WILL get stuck". probably good to clarify that this is to help in the unexpected case that you somehow get stuck.

docs/user_guide/using-submitter.md
eherozhao: this user guide is not only for SMF customers so we would need to include all cases.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+   - You can also specify multi-frame rendering with your job submission. If you do, you can also specify the max percentage of CPU usage you wish to allocate towards rendering in case you would like to limit it to allow other background applications or processes to run smoothly. For more information about multi-frame rendering, visit Adobe's website [here](https://helpx.adobe.com/after-effects/using/multi-frame-rendering.html).
+   - (Optional) You can configure timeout settings (days, hours, and minutes) to handle unexpected cases where a task may become unresponsive. The default timeout is 2 days.
+6. If you see a warning popup window with "You are about to run the script contained in file", you can suppress the warning by following the instruction in the popup or the instructions in [installation.md](installation.md) to disable warnings when submitting jobs.
+7. Install any python libraries if prompted and press the **Login** button in the bottom left if you are not logged in.
+8. Set the farm and queue you are submitting to with the **Settings** button, and click **Submit**.
```
npmacl: this seems like it is almost all duplicated from the user install steps, other than launch after effects as admin. I think we would be better saying, "launch as admin, then follow the same steps as user install".

also, I think it might be good to mention why the user should choose to do a system vs user install.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+- **Job Description** - Add optional details about your render job
+- **Priority** - Set job priority for queue management
+- **Initial State** - Control whether the job starts immediately or remains paused
+- **Max Failed Tasks Count** - Maximum number of tasks that can fail before the job is marked as failed
+- **Max Retries Per Task** - Number of times a failed task will be retried
+- **Max Worker Count** - Maximum number of workers that can work on this job simultaneously
+
+### After Effects Specific Settings
+
+Settings specific to After Effects rendering:
```
npmacl, line 56: since we are defining these here, do we want to just reference this section rather than redefining them in the earlier instructions?

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+- OpenType (`.otf`)
+- TrueType (`.ttf`)
+- TrueTypeCollection (`.ttc`) - majority supported
+- [Adobe Fonts](https://fonts.adobe.com/)
+- Windows bitmap fonts (`.fon`) - only supported on Windows machines
+
+**Troubleshooting fonts:**
+If fonts are missing at render time:
+1. Check that they're installed on the system or your user account
+2. Verify they're being included in the job attachments tab in the submitter
```
npmacl: if these are common issues people have, it might be good to more details on how to do these.

1. might be good to have step by step guide for windows / mac, or link to some other documentation you can check.
2. would be good to clarify that you have to click submit, first. a picture would also be good here, showing what it looks like when they are included.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+
+**Adobe Creative Cloud Fonts:**
+Fonts distributed through Adobe Creative Cloud can be made available for all non-Adobe apps on your workstation, or only made available in Adobe apps. Cloud fonts need to be installed for all non-Adobe apps for use with Deadline Cloud.
```
npmacl: I would reword this to be a little more clear, maybe something like: 
"In order for Deadline Cloud to use Fonts distributed through Adobe Creative Cloud, they must be made available for all non-Adobe apps on your workstation".

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+
+To install fonts for non-Adobe apps in Creative Cloud:
+1. Open Adobe Creative Cloud Desktop.
+2. Click "Adobe Fonts" on the account sidebar under "Your plan" to show the Adobe Fonts panel.
+3. Click "Added fonts" on the "Adobe Fonts" sidebar to show your added fonts.
+4. Click "Install family" next to the fonts you would like to make available for non-Adobe apps.
+
+For information about the other submitter tabs, see the [AWS Deadline Cloud guide for using a submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/jobs-using-submitter.html).
```
npmacl: I don't think this is supposed to be under the fonts section?

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Viewing the Job Bundle
+
+To submit a job, the submitter first generates a Job Bundle, and then uses functionality from the Deadline package to submit the Job Bundle to your render farm to run.
```
npmacl: might be nice to link docs on what a job is here.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+/Library/Frameworks/Python.framework/Versions/3.13/bin/python3
+user@7cf34df03377 ~ % where deadline
+/Library/Frameworks/Python.framework/Versions/3.13/bin/deadline
+```
+7. Now if you run `python --version` and `deadline --version`, you should have access to both of the executables and After Effects should too. Retry job submission.
+
+**For Windows:**
+1. Open Command Prompt.
+2. Run `where python`, `where python3`, `where py`, and `where deadline`. The Deadline check will tell you which Python executable you should be adding to your $PATH.
+3. Press Windows + S and search for "Edit the system environment variables". Note this will require admin access. Click "Environment Variables...". Depending on whether Deadline CLI was a user installation or system installation, open the corresponding $PATH variable.
```
npmacl: I wouldn't say $PATH with a $ here since that doesn't appear on the UI. the options are "User variables for <user>" and "System variables"

under there, the user should click on the Variable "Path", then click "Edit", then in the new window that pops up they should click "New" and paste the path.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,172 @@
...
+user@7cf34df03377 ~ % which python3
+/Library/Frameworks/Python.framework/Versions/3.13/bin/python3
+user@7cf34df03377 ~ % where deadline
+/Library/Frameworks/Python.framework/Versions/3.13/bin/deadline
+```
+7. Now if you run `python --version` and `deadline --version`, you should have access to both of the executables and After Effects should too. Retry job submission.
+
+**For Windows:**
+1. Open Command Prompt.
+2. Run `where python`, `where python3`, `where py`, and `where deadline`. The Deadline check will tell you which Python executable you should be adding to your $PATH.
```
npmacl: probably worth noting, `where X` works for command prompt but not for powershell. for powershell the user needs to type `Get-Command X`

mkdocs.yml
```diff
@@ -0,0 +1,36 @@
...
+
+markdown_extensions:
+  - toc:
+      permalink: true
+
+nav:
+  - Home: index.md
+  - Installation: installation.md
+  - Using the Submitter: using-submitter.md
+  - Telemetry: telemetry.md
```
npmacl: was telemetry moved under the user_guide folder? at least for nuke by default it was just under docs/.

Im not sure if it is picked up here without putting it under docs/user_guide

docs/user_guide/installation.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Note: During the installer process, you will choose between User Install or System Install. macOS users will default to User Install, since System Install is currently not supported on macOS. Read only the section below that matches your selection:**
+
+### User Install
+
+1. Download the Deadline Cloud Submitter installer by following [Step 1: Install the Deadline Cloud Submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html#submitter-installation).
+2. Run the installer (no admin required).
+3. Follow the prompts and select the After Effects submitter. The submitter will be installed to these locations by default:
+   - **Windows**: `C:\Users\<user>\DeadlineCloudSubmitter\Submitters\AfterEffects\AE<version>`
+   - **macOS**: `/Users/<user>/DeadlineCloudSubmitter/Submitters/AfterEffects/AE<version>`
```
viknith: Hasn't this logic changed after @jaikish  updated the user install experience? I believe it is a different path now

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,171 @@
...
+
+**Note: Follow the instructions below that match the install type you selected during installation:**
+
+- **User Install**: Installs the submitter to your user directory. The submitter runs as a standalone window (not dockable). No admin privileges required.
+- **System Install** (Windows only): Installs the submitter to the After Effects application directory. The submitter appears as a dockable panel in the Window menu. Requires admin privileges.
+
+### User Install
+
+1. Launch Adobe After Effects.
+2. Open the submitter by clicking **File > Scripts > Run Script File** and navigate to where the `DeadlineCloudSubmitter.jsx` file is located and select it to run the submitter. If you recently closed the submitter, you can reopen it by clicking **File > Scripts > Recent Script Files** and pick the `DeadlineCloudSubmitter.jsx` file that was previously run.
```
viknith, line 20: Same issue as the comment above, the new users don't have to do this anymore they can just follow the original flow of triggering the script when installed via system install

docs/user_guide/installation.md
eherozhao: updated. Wait for his validation now.

docs/user_guide/installation.md
```diff
@@ -0,0 +1,86 @@
...
+     The submitter will appear in the **Window** menu in After Effects.
+   - **Windows**: The submitter is installed to a standalone location by default:
+     ```
+     C:\Users\<user>\DeadlineCloudSubmitter\Submitters\AfterEffects\AE<version>
+     ```
+     You will need to run the script manually via **File > Scripts > Run Script File**.
+
+### System Install
+
+**Note: macOS system install is not yet supported by the submitter installer. macOS users must do User Install.**
```
xuetinp: If user has admin access on MacOS, they can directly place the files under Applications/Adobe After Effects \<version\>/Scripts/ScriptUI Panels just like admin installation on windows right?

docs/user_guide/installation.md
```diff
@@ -0,0 +1,86 @@
...
+   To allow scripts to write files or send communication over a network:
+   
+   - **Windows**: Select `Edit > Preferences > Scripting & Expressions > select Allow Scripts To Write Files And Access Network`
+   - **macOS**: Select `After Effects > Settings > Scripting & Expressions > select Allow Scripts To Write Files And Access Network`
+
+## Installing the submitter
+
+The After Effects submitter extension allows you to submit jobs to Deadline Cloud directly from within After Effects.
+
+**Note: During the installer process, you will choose between User Install or System Install. macOS users will default to User Install, since System Install is currently not supported on macOS. Read only the section below that matches your selection:**
```
xuetinp: >since System Install is currently not supported on macOS

Should we clarify it's automatic system install since manual system install is supported

docs/user_guide/installation.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. Locate the `DeadlineCloudSubmitter.jsx` file and the `DeadlineCloudSubmitter_Assets` folder in the `dist` folder of this repository.
+2. Copy both to the **ScriptUI Panels** folder within your After Effects installation:
+   - **Windows**: `Program Files\Adobe\Adobe After Effects <version>\Support Files\Scripts\Script UI Panels`
+   - **macOS**: `Applications/Adobe After Effects <version>/Scripts/Script UI Panels`
+3. Restart After Effects if it was open.
+
+## Setting up After Effects with your Deadline Cloud Farm
+
+After Effects 24.6, 25.1, and 25.2 conda packages are available in AWS Deadline Cloud Service Managed Fleet (See this [link](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/create-queue-environment.html) for more information). If you would like to build a conda channel that contains a different After Effects conda package, please follow [these instructions](https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/configure-jobs-s3-channel.html).
```
xuetinp: Maybe we don't mention specific AE versions but provide a link for customer to check, so when we support newer versions or deprecate old versions, we don't need to modify this file

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,179 @@
+# Using the After Effects Submitter
+
+To use the Deadline Cloud submitter for After Effects, please ensure your Farm is configured with an After Effects capable fleet, and have the submitter installed. Also, please log into the Deadline Cloud Monitor or provide AWS credentials via a configuration profile for Deadline Cloud access.
```
xuetinp: >please ensure your Farm is configured with an After Effects capable fleet

Could we provide a link to explain AE capable fleet? If we don't have such info, maybe elaborate a bit more here.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,179 @@
...
+Refer to the [installation.md](installation.md) to configure the fleet and After Effects application.
+
+## Submit a job
+
+**Note: Follow the instructions below that match the install type you selected during installation:**
+
+- **User Install**: Installs the submitter to your user directory. No admin privileges required.
+  - **macOS**: The submitter is automatically installed to your After Effects user preferences directory and appears in the **Window** menu as `DeadlineCloudSubmitter(User).jsx`.
+  - **Windows**: The submitter is installed as a standalone script that you run via **File > Scripts > Run Script File**.
+- **System Install** (Windows only): Installs the submitter to the After Effects application directory. The submitter appears as a dockable panel in the **Window** menu. Requires admin privileges.
```
xuetinp: The summary here seems redundant from the detailed instructions below. I think it's easier for people to just follow the steps for their install type.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,179 @@
...
+7. Install any python libraries if prompted and press the **Login** button in the bottom left if you are not logged in.
+8. Set the farm and queue you are submitting to with the **Settings** button, and click **Submit**.
+
+### User Install - Windows
+
+1. Launch Adobe After Effects.
+2. Open the submitter by clicking **File > Scripts > Run Script File** and navigate to where the `DeadlineCloudSubmitter.jsx` file is located and select it to run the submitter. If you recently closed the submitter, you can reopen it by clicking **File > Scripts > Recent Script Files** and pick the `DeadlineCloudSubmitter.jsx` file that was previously run.
+3. Follow steps 3-8 from the [User Install - macOS](#user-install---macos) section above.
+
+### System Install (Windows only)
```
xuetinp: System Install is possible for MacOS admin users as well. The steps are the same as system install for windows. We may just remove `(Windows only)` in the title.

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,179 @@
...
+6. If you see a warning popup window with "You are about to run the script contained in file", you can suppress the warning by following the instruction in the popup or the instructions in [installation.md](installation.md) to disable warnings when submitting jobs.
+7. Install any python libraries if prompted and press the **Login** button in the bottom left if you are not logged in.
+8. Set the farm and queue you are submitting to with the **Settings** button, and click **Submit**.
+
+### User Install - Windows
+
+1. Launch Adobe After Effects.
+2. Open the submitter by clicking **File > Scripts > Run Script File** and navigate to where the `DeadlineCloudSubmitter.jsx` file is located and select it to run the submitter. If you recently closed the submitter, you can reopen it by clicking **File > Scripts > Recent Script Files** and pick the `DeadlineCloudSubmitter.jsx` file that was previously run.
+3. Follow steps 3-8 from the [User Install - macOS](#user-install---macos) section above.
```
jaikish: this is pointing to MacOS in Windows section

eherozhao: yes, since the usage of the tool will be the same no matter which os they are in.

eherozhao: Updated the doc to make it easy to follow.

jaikish (approved):

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,178 @@
+# Using the After Effects Submitter
+
+To use the Deadline Cloud submitter for After Effects, please ensure your Farm is configured with an After Effects capable fleet (see [Setting up After Effects with your Deadline Cloud Farm](installation.md#setting-up-after-effects-with-your-deadline-cloud-farm)), and have the submitter installed. Also, please log into the Deadline Cloud Monitor or provide AWS credentials via a configuration profile for Deadline Cloud access.
```
xuetinp, line 3: The link points to a 404 page for me

eherozhao: this only worked in the rendered view. check this https://github.com/eherozhao/deadline-cloud-for-after-effects/blob/mainline/docs/user_guide/using-submitter.md

xuetinp (approved):

PR #291, opened 2026-03-06
feat: add AE 2026 support to installer and submitter

viknith:
We need AE 26 and 25.6 installer support

Added it

We are supporting AE 26.

Did manual testing of the new installer in Windows and macOS

No

#### If you modified the `/installer`, `/install_builder`, or `/scripts` logic, please run the integration tests and paste the results below=

Mac
```
======================== 4 passed, 9 skipped in 12.91s =========================
```
Windows user install test
```
===================== 4 passed, 9 skipped in 31.03s ===============
```
Windows system install test
```
========== 7 passed, 6 skipped in 30.22s ==========================
```

Yes, but another update will be needed later once AE 26 conda is shipped

No

---


README.md
homjason, line 9: nit: Should we update this to `2024-2026`?

homjason (approved):

README.md
viknith: I thought about it, and decided not to do it until we actually do the conda release because otherwise people might be confused

eherozhao (approved):

PR #298, opened 2026-04-20
chore: process improvements

crowecawcaw:
Release process is a bit tedious.

- Improve pip and hatch caching with CI actions. (Caching was added in shared workflows already.)
- Add permissions to release note process so they can be LLM generated
- Add a single manual approval step instead of requiring manual approvals on every release step. Before merging, will require a new ManualReleaseGate environment and removing approvals for the release environment

Faster releases

Works in deadline-cloud repo

n/a

No

----

### DIFF
--- a/.github/workflows/release_publish.yml
+++ b/.github/workflows/release_publish.yml
@@ -70,7 +70,7 @@ jobs:
       project_name: ${{ github.event.repository.name }}
           
   Release:
-    needs: [TagRelease, ManualTestGate]
+    needs: [TagRelease, ReleaseInstaller]
     runs-on: ubuntu-latest
     name: Release
     permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/release_publish.yml
```diff
@@ -63,7 +70,7 @@ jobs:
       project_name: ${{ github.event.repository.name }}
           
   Release:
-    needs: [TagRelease, ReleaseInstaller]
+    needs: [TagRelease, ManualTestGate]
```
jaikish: Shouldn't we still have ReleaseInstaller in `needs`?

crowecawcaw: Fixed!

viknith (approved):

viknith (approved):

jaikish (approved):

jaikish (approved):

PR #300, opened 2026-04-28
docs: Add ae-dev-setup Kiro skill and FAQ doc

xuetinp:

The After Effects integration repo lacked:
1. A Kiro skill for automated dev environment setup 
2. A FAQ document for troubleshooting common job failures

**Kiro skill (`skills/ae-dev-setup/`):**
- `SKILL.md` — skill metadata, agent behavior rules, and workflow summary
- `references/setup-guide.md` — step-by-step agent workflow: OS detection (stops on Linux), Python verification, AE detection, Hatch install, dependency sync, submitter build via jsxbundler, and optional copy to ScriptUI Panels

**FAQ (`docs/user_guide/faq.md`):**
- Custom ExtendScript on workers — timeout and PermissionDenied errors caused by UI dialogs in headless mode
- "aerender Error: Could not read from source" — missing output path in render queue
- Font install failure — unsupported `.ttc` fonts on Windows workers

No impact on existing functionality. Adds documentation only.

Verified markdown renders correctly. No source code changes.

No — no source files were modified.

N/A

N/A

Yes — this change is documentation itself.

No.

---


docs/user_guide/faq.md
```diff
@@ -0,0 +1,38 @@
+# Frequently Asked Questions
+
+## Rendering Questions
+
+**Q: Can I run custom ExtendScript (.jsx) scripts on Deadline Cloud workers?**
+
+A: Yes. The After Effects conda package on Service Managed Fleets includes the full After Effects application, not a stripped-down version. You can run custom `.jsx` scripts on workers using host configurations.
```
viknith: Note: There are some limitations though. On our official Conda package, we do package AfterFX.exe which helps run scripts. However the scripts do not work on worker if they ever make networking calls, because that is a user setting that an artist would normally set locally and it cannot be set on the worker because we don't do an actual AE install on the worker and we just download the renderer and its dependencies instead.

docs/user_guide/faq.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Troubleshooting tip:** Try running the same workflow on a Windows workstation with the After Effects UI to eliminate any non-Deadline-specific issues.
+
+**Q: I'm getting "aerender Error: Could not read from source" when my job runs. What's wrong?**
+
+A: This error typically means the output path was not set in your composition's render queue before submitting the job. The After Effects submitter requires that you add your composition to the render queue and set up your render settings, output module, and output path before submission. See [Using the After Effects Submitter](using-submitter.md) for the full submission workflow.
+
+**Q: My job fails at "Install Fonts to Worker" with `AddFontResource failed to load` on a `.ttc` font. What do I do?**
+
+A: Not all TrueTypeCollection (`.ttc`) fonts are installable on Windows. When the font manager attempts to install an unsupported `.ttc` font on a worker, it fails with an error like:
```
viknith: I think we discussed this before and its not just a limitation with .ttc and the issue can also be related to any font file as long as there is a limitation with OS specific stuff.

skills/ae-dev-setup/SKILL.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Agent Behavior
+
+You **MUST** automate all possible steps by running commands directly.
+You **MUST** only prompt the user when their input is required (AE version, custom paths).
+You **MUST** validate each step's output before proceeding to the next.
+You **MUST** inform the user clearly when manual intervention is needed (e.g., installing After Effects itself).
+You **SHOULD** report progress at each step.
+
+## Setup Workflow
```
viknith, line 34: Nit: do we need a separate markdown under references if this is the only skill using that setup guide?

docs/user_guide/faq.md
xuetinp: Sure, I can make this question more font-agnostic.

docs/user_guide/faq.md
xuetinp: Will update

skills/ae-dev-setup/SKILL.md
xuetinp: This is the style we follow in other deadline cloud repos. I prefer to keep it consistent.

docs/user_guide/faq.md
```diff
@@ -4,7 +4,7 @@
 
 **Q: Can I run custom ExtendScript (.jsx) scripts on Deadline Cloud workers?**
 
-A: Yes. The After Effects conda package on Service Managed Fleets includes the full After Effects application, not a stripped-down version. You can run custom `.jsx` scripts on workers using host configurations.
+A: Yes, with limitations. The After Effects conda package on Service Managed Fleets includes the full After Effects application, not a stripped-down version. You can run custom `.jsx` scripts on workers using host configurations. However, scripts that make networking calls will not work on workers, because the "Allow Scripts To Write Files And Access Network" preference is a local user setting that cannot be configured on workers since After Effects is not fully installed — only the renderer and its dependencies are packaged.
```
viknith: Nit: the first sentence ` The After Effects conda package on Service Managed Fleets includes the full After Effects application` contradicts the last sentence which says `After Effects is not fully installed — only the renderer and its dependencies are packaged`, which implies its a stripped down version. We should just say its a stripped down version that only installs the renderer and its dependencies.

xuetinp: Good catch! Fixed

viknith (dismissed):

docs/user_guide/faq.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Q: My job fails at "Install Fonts to Worker" with `AddFontResource failed to load`. What do I do?**
+
+A: Some fonts are not installable on certain operating systems due to OS-specific limitations. When the font manager attempts to install an unsupported font on a worker, it fails with an error like:
+```
+OSError: AddFontResource failed to load "...\tempFonts\HelveticaNeue-CondensedBlack.ttc"
+```
+
+To resolve this:
+1. Check whether the font is actually used in your composition — it may have been uploaded by mistake.
```
jaikish, line 33: One more option is creating shapes from text if this is acceptable for customers. 

`Create shapes from text characters`: https://helpx.adobe.com/after-effects/using/creating-shapes-masks.html

The downside is : once it is converted, they cannot edit the text and rely on text specific features

xuetinp: Updated

jaikish (approved):

viknith (approved):

PR #323, opened 2026-08-04
fix: avoid Terminal cold-start race on macOS submission

### LINKED ISSUE #305: Bug: On mac, submitter terminal command sometimes doesn't run.
### Describe the bug

Image

On my mac, it sometimes takes up to 5 seconds for a terminal window to load. The terminal window will appear, but it remains blank for a while with no shell running. This seems to create a race condition where, the terminal sometimes tries to process the command before the shell starts. My screenshot shows the command appearing before fish's welcome message.

### Expected Behaviour

The command should be run every time.

### Current Behaviour

Submissions sometimes silently fail to bring up the submitter window.

### Reproduction Steps

install the submitter on mac, quit the terminal app, try to submit a job.

### Environment

At minimum:
1. Operating system: macOS Tahoe 26.5.1
2. Version of After Effects: 2026
3. Version of this package: ???

jaikish:
## Problem

Fixes #305.

The submitter launched the Deadline GUI by injecting a command into Terminal.app via AppleScript `do script`. This had two problems:

1. **Silent drop on cold start.** Terminal takes several seconds to launch its login shell. A `do script` sent before the shell is ready is silently dropped — the window opens blank and no GUI appears. The `busy` property fires when the shell *process* has started, not when it is interactive. With a slow-loading shell config (e.g. fish with plugins) this gap can be several seconds.

2. **After Effects UI blocks.** `system.callSystem` is synchronous. Any delay during deadline startup blocked the entire AE application.

## Fix

Replace Terminal entirely on macOS with a Python script, `launch_deadline.py`.

- `/usr/bin/python3` is always on the minimal launchd PATH that After Effects inherits from Finder — no shell profile sourcing needed
- `launch_deadline.py` finds the `deadline` binary by its well-known installer path (`~/DeadlineCloudSubmitter/DeadlineClient/deadline`) and **double-forks** before launching `gui-submit`
- The fork calls `os.setsid()` to start a new session, detaching the deadline process from After Effects' process group. `python3` exits immediately after the fork, so `system.callSystem` returns in ~64ms regardless of what deadline does afterward
- Falls back to `shutil.which("deadline")` for pip-installed setups

The Windows branch is unchanged. Its `ERROR CODE` error check is preserved inside the Windows branch.

## Changes

- `src/scripts/launch_deadline.py` — new Python launcher (mirrored to `dist/…/scripts/`)
- `src/submission/SubmitBundle.jsx` — macOS `else` branch replaced with `launchDeadlineGUI()`; Terminal/osascript machinery removed
- `launch_deadline.py` added to the required-files check

## Notes on automated review comments

**Comment on `singleQuoteForShell` / path injection:** Not applicable — that construction was removed by this PR. The bundle path is passed as a Python `sys.argv` argument via `subprocess.Popen` with no shell involved.

**Comment on `--submitter-info` vs `--submitter-name`:** `--submitter-info submitter_name=After Effects` is the correct current syntax. The CLI marks `--submitter-name` as deprecated in favour of `--submitter-info`.

**Comment on `ERROR CODE` check removed:** Fixed — the check is now scoped inside the Windows branch where `output` is populated.

## Testing

- Bug reproduced: cold Terminal + slow fish login shell drops 5/5 without a wait and 1/5 with the previous `busy`-wait approach
- Fix passes 5/5 cold-start trials; GUI confirmed launching and making real AWS API calls
- AE UI confirmed non-blocking: `system.callSystem` returns in ~64ms
- 7/7 unit tests pass
- Bundle regenerated via `jsxbundler.py`

### DIFF
--- a/dist/DeadlineCloudSubmitter.jsx
+++ b/dist/DeadlineCloudSubmitter.jsx
@@ -2109,6 +2109,29 @@ function generateJobEnvironmentFragment(bundlePath, outputFoldersStr) {
 }
 
 /**
+ * Launches the Deadline Cloud GUI submitter on macOS without using Terminal.
+ *
+ * After Effects inherits a minimal launchd PATH from Finder that does not
+ * include the deadline CLI. Routing through Terminal.app to source the user's
+ * shell profile created a race: on a cold start the shell may not be ready
+ * before `do script` fires, silently dropping the command.
+ *
+ * Instead, system python3 (/usr/bin/python3) — which is always on the launchd
+ * PATH — runs launch_deadline.py, which locates the deadline binary by its
+ * well-known installer path and launches gui-submit as a detached process.
+ *
+ * @param {String} bundlePath         path to the job bundle directory
+ * @param {String} pythonExecutable   python3 executable found by getPythonExecutable()
+ **/
+function launchDeadlineGUI(bundlePath, pythonExecutable) {
+    const launcherScript = scriptFolder +
+        "/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/launch_deadline.py";
+    const cmd = pythonExecutable + " \"" + launcherScript + "\" \"" + bundlePath + "\"";
+    logger.debug("Launching Deadline GUI via Python launcher: " + cmd, submitBundleFile);
+    system.callSystem(cmd);
+}
+
+/**
  * Submit the selected render queue item
  **/
 function SubmitSelection(selection, selectionSettings) {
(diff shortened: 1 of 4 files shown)


yuanmich2:
Just leaving a comment here to say that this PR unfortunately doesn't fix the issue for me.

yuanmich2:
Sorry! I thought I posted again yesterday, but I don't see my comment. It turns out my testing method was flawed because the submitter install failed and I was actually using an old version of the call_aerender.py script. I need to wipe my AE config and retest

yuanmich2:
Just finished testing and I really like the changes. A Terminal window no longer pops up and the submitter appears reliably for me.

